PR #1, opened 2015-04-01
Change to use browserify and dev/dist build folders with gulp flags

dougmartin:
This changes build system to use browserify with a set of gulp flags:

    --production  
    --buildInfo <string>

If --production is set the output folder is ./dist otherwise it is ./dev (formerly public).  The production flag also controls if stylus compresses the output css.

The --buildInfo string is used to replace the marker in the build-info meta tag in index.html.  If it is not present the string 'development build ([date])' is inserted, with [date] built from (new Date()).toString().

The build.sh script has been updated to use the new flags.

Here is an example gulp run during development

    gulp

and here is what is used in build.sh

    gulp build-all --production --buildInfo '$SHA built on $DATE'

NOTES:

1. You will need to run 'npm install' to get the new modules needed
2. This pull request DOES NOT change the .jsx files into CoffeeScript.  I thought best if that was done in a separate request if needed after this one was accepted for clarity.
3. The touchPunch, react, lodash and loglevel modules have been placed in a separate 'globals.coffee' file (maps to /js/globals.js).  This reduces build and reloading time during development.
4. The .coffee and .jsx extensions are needed in the require statements for now since we have .jsx files.  If they are converted then we can drop the extensions.


dougmartin:
I added and rebased a commit to 1) update the README.md to remove the jspm references 2) add source-maps if the production flag is NOT set and 3) fixup the Mocha tests

knowuh:
This looks great *and* removes 627 lines of noise.

I Easily built and ran in dev / and production.  I didn't take a good look in the dev console to see that the network requests all made sense, but I assume they do.

The Mocha tests are failing. But that may or may not be a related regression.

dougmartin:
The Mocha tests were failing before the code change - I just updated them to point to the correct files.

PR #2, opened 2015-04-02
Moved code to CoffeeScript

dougmartin:
This is a pretty big pull request.  It encompasses the following:

1. Renaming the 'javascripts' folder to 'code' and creating subfolders for views and utils
2. Converting all the .jsx files to .coffee files and removing the reactify gulp step
3. Converting all the .js files to .coffee (except for the touchpunch.js vendor file)
4. Fixing up the Mocha tests to run
5. Copying down the remote images and putting them in the assets folder
6. Removing unneeded dependencies that have accumulated in package.json
7. Tweaking some of the existing CoffeeScript code to make it more idiomatic
8. Removing a couple of dead util files not used anywhere in the system
9. Removing the .coffee extension in the requires

This pull request DOES NOT include any linting of the coffee files.  The style linter is a little too picky for my tastes (especially on max line lengths).  I'll play around more with the settings in a separate pull request once this one is merged.


knowuh:
Sorry Doug, I didn't get to this this afternoon.  I am going to have a look first thing in the AM.

Cheers,  - n

src/code/models/link-manager.coffee
```diff
@@ -30,17 +30,17 @@ class LinkManager
     log.info("adding selection listener #{listener}")
     @selectionListeners.push listener
 
-  getLinks: () ->
-    return (value for key, value of @linkKeys)
```
knowuh: Not a fan of explicit returns?  I am inconsistent on it, which is far worse.

src/code/views/link-view.coffee
```diff
@@ -0,0 +1,148 @@
...
+            selected: node.selected
+            nodeKey: node.key
+            ref: node.key
+            onMove: @onNodeMoved
+            onDelete: @onNodeDeleted
+            linkManager: @props.linkManager
+          })
+      )
+    )
+
```
knowuh: So clean. Thanks.

src/code/views/app-view.coffee
```diff
@@ -0,0 +1,67 @@
+LinkView    = React.createFactory require './link-view'
+NodeWell    = React.createFactory require './node-well-view'
+NodeEditView= React.createFactory require './node-edit-view'
+LinkEditView= React.createFactory require './link-edit-view'
+StatusMenu  = React.createFactory require './status-menu-view'
```
knowuh: This is new to me.  If you have the time I would love to understand that 'createFactory' does in this case.

src/code/models/link-manager.coffee
dougmartin: The link-manager.coffee file is the one existing coffee file I didn't finish cleaning up.  I noticed it when I was reviewing the pull request but I went ahead and pulled the trigger.  I'll clean it up when I add a coffeelint gulp task.

knowuh:
Looks great.  Thanks for the improvements @dougmartin 

 As far as including the images as part of the project, I understand why we have done it.  We probably need to think about finding images which we have the rights to use though. (TBD w/ @ddamelin )  - much of this project was thrown together as a technology spike.

In day to day use, we expect students and authors we mostly be using an image library that we provide, however we want folks to be able to point to images using any URL scheme.

Tests pass, & build works fine for me.  Feel free to update the github pages demo site using build.sh whenever you think you have a working build in your process that you want to share.

I had one question about using React.createFactory in a require, but I could google that on my own.

src/code/views/app-view.coffee
dougmartin: This was new in React 0.12 and is related to them moving to a pure class based system.  

A good explanation is here: https://facebook.github.io/react/blog/2014/10/14/introducing-react-elements.html

In 0.13 passing a function generated by createClass in another components render method works but generates a console warning.  I added the createFactory calls to suppress the warning messages.  The JSX compiler automatically adds the createFactory() wrapper which is why you didn't see the warnings before.

dougmartin:
Great - I'll update the demo site today.  I pulled the images down because one of them was hotlinked off a site and you never know when someone might notice and switch the hotlinked image to something *more interesting*.

PR #5, opened 2015-04-03
Added .travis.yml and updated package.json

dougmartin:
Adds the travis file and the missing test script and coffee-script dependency to package.json (so that mocha can run out of node_modules)


knowuh:
Thanks.

PR #7, opened 2015-04-06
Changed node edit to use select for image

dougmartin:
Changes the node edit form to use a select for the image type.

Closes https://www.pivotaltracker.com/story/show/89739108


knowuh:
Nice.  Thanks.

PR #8, opened 2015-04-07
Adds a remote image option to the node edit view

dougmartin:
The node edit image select now has two option groups - built-in images and remote images.  Under the remote images option group there is an 'Add Remote...' option that when selected shows a remote url input and a button to add the image.  When the add image button is clicked the img url is tested to see if it can be loaded and if it can the image is changed and is also added to the palette to the left of the blank option.

I also added the ability to click on an image in the palette and immediately update the selected node to that image.  This was not in the story but I thought it would be interesting to test.

### DIFF
--- a/src/code/views/app-view.coffee
+++ b/src/code/views/app-view.coffee
@@ -8,26 +8,6 @@ StatusMenu  = React.createFactory require './status-menu-view'
 
 log.setLevel log.levels.TRACE
 
-
-protoNodes = [
-  {
-    'title': 'Egg',
-    'image': 'img/nodes/egg.png'
-  },
-  {
-    'title': 'Chick'
-    'image': 'img/nodes/chick.jpg'
-  },
-  {
-    'title': 'Chicken'
-    'image': 'img/nodes/chicken.jpg'
-  },
-  {
-    'title': ''
-    'image': ''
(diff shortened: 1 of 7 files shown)


src/code/views/app-view.coffee
```diff
@@ -74,14 +55,30 @@ module.exports = React.createClass
 
   onLinkChanged: (link, title, color, deleted) ->
     @props.linkManager.changeLink title,color, deleted
-  
+
+  onAddRemoteImage: (image) ->
+    # make sure this is a new image
+    if not _.find @state.protoNodes, {image: image}
```
knowuh: Do you think it might time to refactor?  I suspect that this logic could be more lucidly expressed in a thicker proto-node (or proto-node-list) implementation.

src/code/views/node-edit-view.coffee
```diff
@@ -10,21 +10,53 @@ module.exports = React.createClass
   changeImage: (e) ->
     @props.onNodeChanged? @props.node, @props.node.title, e.target.value
 
+  addRemote: (e) ->
```
knowuh: Not sure, but this also seems like a concern that could get pushed out of the view.

PR #9, opened 2015-04-08
Added new button and google drive open and save buttons

dougmartin:
Along with a new button and google drive open and save buttons this also:

1. Adds a version number, the filename and the state of the node well to the serialized data.  The version number can be used in the future and the filename is used when serializing to the url.  
2. Adds a load listener to the link manager that is used in the app view to both load the saved palette and replace the onAddRemoteImage() functionality 
3. Moves the filename into the link manager and adds a listener that the app view subscribes to so that it can update its filename state which is passed down to the google file view
4. Removes the node well click handler that updates the selected node image

### DIFF
--- a/src/assets/json/serialized.json
+++ b/src/assets/json/serialized.json
@@ -1,4 +1,6 @@
 {
+    "version": 0.1,
+    "filename": "Sample Model",
     "nodes": [
         {
             "title": "Chicken",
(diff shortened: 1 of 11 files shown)


dougmartin:
I went back and forth about keeping the filename in the link manager.  I think the link manager needs to become a mixin for the app-view to make it more 'Reacty'.

knowuh:
Next week I might get you to talk me through setting up google drive API for testing those features in my local dev environment.

PR #11, opened 2015-04-14
Added drag and drop support for local images

dougmartin:


src/code/views/node-edit-view.coffee
```diff
@@ -22,6 +22,17 @@ module.exports = React.createClass
       img.src = src
 
   render: ->
+    builtInNodes = []
+    droppedNodes = []
+    remoteNodes = []
+    for node, i in @props.protoNodes
+      if not node.image.match /^(https?|data):/
```
knowuh: Could be wrong, but seems like first case should only match `/^https?/` ? Otherwise it seems like next case will never match.

PR #20, opened 2015-04-28
Added image browser

dougmartin:
This is the image browser minus the metadata viewing.  We need to restructure how the metadata is stored.  I will also add the remaining new internationalized strings missing in this commit when the metadata viewing is added.

### DIFF
--- a/src/code/mixins/app-view.coffee
+++ b/src/code/mixins/app-view.coffee
@@ -20,27 +20,25 @@ module.exports =
           deleteFunction()
     else
       $(window).off 'keydown'
+      
+  addToPalette: (node) ->
+    if node?.image.match /^(https?|data):/
+      # make sure this is a new image
+      if not _.find @state.protoNodes, {image: node.image}
+        protoNodes = @state.protoNodes.slice 0
+        protoNodes.push
+          title: node.title or ''
+          image: node.image
+        @setState protoNodes: protoNodes
 
   componentDidMount: ->
     @addDeleteKeyHandler true
-
-    updatePalette = (node) =>
-      if node?.image.match /^(https?|data):/
-        # make sure this is a new image
-        if not _.find @state.protoNodes, {image: node.image}
-          # add the image before the empty image
-          protoNodes = @state.protoNodes.slice 0
-          emptyPos = _.findIndex protoNodes, {image: ''}
-          protoNodes.splice (if emptyPos is -1 then protoNodes.length else emptyPos), 0,
-            title: node.title or ''
-            image: node.image
-          @setState protoNodes: protoNodes
 
     @props.linkManager.addSelectionListener (selections) =>
       @setState
         selectedNode: selections.node
         selectedConnection: selections.connection
-      updatePalette selections.node
+      @addToPalette selections.node
       log.info 'updated selections: + selections'
 
     @props.linkManager.addLoadListener (data) =>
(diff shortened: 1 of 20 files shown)


src/code/mixins/app-view.coffee
```diff
@@ -20,27 +20,25 @@ module.exports =
           deleteFunction()
     else
       $(window).off 'keydown'
+      
+  addToPalette: (node) ->
+    if node?.image.match /^(https?|data):/
+      # make sure this is a new image
+      if not _.find @state.protoNodes, {image: node.image}
```
knowuh: Do we need a Palette model here so that the view doesn't have to interrogate the data so hard?  This Palette model could have an "addToPalette" method which could do these checks and handle ordering &etc.

src/code/views/image-browser-view.coffee
```diff
@@ -0,0 +1,172 @@
...
+    image.onload = =>
+      @setState loaded: true
+  clicked: ->
+    @props.clicked @props.imageInfo
+  render: ->
+    src = if @state.loaded then @props.imageInfo.image else 'img/bb-chrome/spin.svg'
+    (img {src: src, onClick: @clicked, title: @props.imageInfo.title})
+
+PreviewImage = React.createFactory React.createClass
+  displayName: 'ImageSearchResult'
```
knowuh: displayName Should probably be 'PreviewImage'

src/code/views/image-browser-view.coffee
```diff
@@ -0,0 +1,172 @@
...
+
+  imageClicked: (imageInfo) ->
+    @setState selectedImage: imageInfo
+
+  addImage: (imageInfo) ->
+    if imageInfo
+      @props.addToPalette imageInfo
+    @setState selectedImage: null
+    
+  render: ->
```
knowuh: This render method seems a bit long in the tooth.  Any thoughts for how to simplify?

src/code/views/modal-tabbed-dialog-view.coffee
```diff
@@ -0,0 +1,43 @@
...
+    (li {className: (if @props.selected then 'tab-selected' else ''), onClick: @clicked}, @props.label)
+
+module.exports = React.createClass
+
+  displayName: 'ModalTabbedDialog'
+
+  getInitialState: ->
+    selectedTabIndex: 0
+
+  statics:
```
knowuh: Here I learned something about react component `statics`.

dougmartin: I probably shouldn't have overloaded the word 'Tab'.  The render method uses the Tab from line 8.  The Tab on line 27 lets us export the TabInfo class.

src/code/views/image-browser-view.coffee
```diff
@@ -0,0 +1,172 @@
...
+Link = React.createFactory React.createClass
+  displayName: 'Link'
+  render: ->
+    (div {}, 'Link: TBD')
+
+module.exports = React.createClass
+  displayName: 'Image Browser'
+  render: ->
+    (ModalTabbedDialogFactory {title: (tr "~ADD-NEW-IMAGE.TITLE"), close: @props.close, tabs: [
+      ModalTabbedDialog.Tab {label: (tr "~ADD-NEW-IMAGE.IMAGE-SEARCH-TAB"), component: (ImageSearch {protoNodes: @props.protoNodes, addToPalette: @props.addToPalette})}
```
dougmartin: You can see the use of the Tab static function here to create an object to describe a tab.

PR #23, opened 2015-04-30
Ongoing changes to image browser

dougmartin:
Removed dynamic searching, changed search to sort by download, added metadata saving and viewing and translation strings.

NOTE: I added a simple variable substitution system for translations strings.


knowuh:
Looks good. Thanks for adding vars to the strings templates.  

One general question I had was should the image meta-data be part of the Prototype Node data structure, instead of in a lookup table?  I didn't see a case where we would need the meta-data outside of the context of the ProtoNode.

PR #24, opened 2015-05-01
Added undo/redo manager

dougmartin:
I could not find styling for the 'Unsaved' notice in the global nav bar or the undo/redo buttons in the document actions so I made placeholder elements for now.


dougmartin:
@knowuh Please review when you get a chance.  The link manager diff is a little ugly since I had to move some methods around.

knowuh:
Looks good.  We might want to write some tests for some of this, since it shouldn't be too dificult.  I like how the node dragging is debounced; we might want to debounce updates to text labels in the future.

PR #25, opened 2015-05-01
Node editing palette

knowuh:
@dougmartin can you take a look?


src/code/models/link-manager.coffee
```diff
@@ -185,21 +185,28 @@ module.exports = class LinkManager
...
-  changeNode: (title, image) ->
+  changeNode: (data) ->
     if @selectedNode
       node = @selectedNode
-      originalTitle = @selectedNode.title
-      originalImage = @selectedNode.image
+      originalData =
+        title: node.title
+        image: node.image
+        color: node.color
```
dougmartin: I wonder if it would be better to use _.cloneDeep when saving this reference so that any new data members automatically get copied.

    originalData = _.cloneDeep data

PR #26, opened 2015-05-01
Fix item drops on inspector panel [#93637418]

dougmartin:


dougmartin:
@knowuh This is a quick one.

PR #28, opened 2015-05-05
Adds internal image library and disables image search results already in

dougmartin:
palette [#93795450] [#93633012].


dougmartin:
@knowuh Please review.  It is a lot of renaming of protoNodes to palette and then splits the palette and the internal library into two lists.

src/code/views/image-browser-view.coffee
```diff
@@ -125,9 +128,9 @@ ImageSearch = React.createFactory React.createClass
             if @state.internalResults.length is 0 and (@state.searching or @state.externalResults.length > 0)
```
knowuh: This part set of conditional blocks (L128 - L133) is hard for me to reason about, because the intent isn't obvious. Any thoughts on how to make it more obvious?

knowuh:
Looks good.

PR #29, opened 2015-05-05
Fix warnings about React collection unique key requirements.

knowuh:
This adds required key to missing collections. (mostly because I was tired of the warnings)


knowuh:
@dougmartin  just some tiny changes here.

dougmartin:
Looks good.

PR #33, opened 2015-05-06
Fixed images already in the palette should be disabled  [#93915252]

dougmartin:


dougmartin:
I'm going to merge this myself - it is very small.

PR #34, opened 2015-05-06
Fixes clicking open when have two logged in users in Chrome [#93911702]

dougmartin:
The auto token is now saved in case the user picks a new user to authorize in the popup dialog.


dougmartin:
I'm going to merge this myself as it is small and so I can rebase the refactor-stylus-fonts PR.

PR #36, opened 2015-05-06
Added revert to original and revert to last save menu options [#93911902]

dougmartin:
I also converted the strings in google-file-interface to use tr() and changed the styling of inactive menu options.


dougmartin:
@knowuh Please review.

PR #37, opened 2015-05-06
Added editable image credits for externally dropped images [#92310054]

dougmartin:


dougmartin:
@knowuh Please review.

src/code/data/internal-library.coffee
```diff
@@ -33,6 +41,7 @@ module.exports = [
     "title": "Cloud"
     "image": "img/nodes/cloud.png",
     "metadata": {
+      "source": "internal",
       "title": "Cloud",
       "link": "https://openclipart.org/detail/17666/net wan cloud"
```
knowuh: "net wan cloud"  seems like a weird URL.

src/code/mixins/app-view.coffee
```diff
@@ -1,11 +1,17 @@
 module.exports =
 
   getInitialAppViewState: (subState) ->
+
+    # load the metadata at startup
+    internalLibrary = require '../data/internal-library'
+    for node in internalLibrary
+      @props.linkManager.setImageMetadata node.image, node.metadata
```
knowuh: You convinced me once that it was a good idea manage the image metadata outside of the node.  But I am renewing my doubts about this particular approach. I assume the motivation is to speed up the lookup of meta-data, otherwise we could do an iteration search over the library and palette items.

(de) serializing the metadata as part of the node data also seems like it would be logical, and make the node types more atomic.

What do you think about developing a class to manage the combined library / palette collections? I had thought of of the link-manager as managing the state of the graph -- Its growing in scope, and I want to keep that in check if possible, or break it apart.

PR #38, opened 2015-05-07
Drag from one source, to anywhere in node. [#93635084]

knowuh:
https://www.pivotaltracker.com/story/show/93635084

Improve link style, make nodes look like PSDs.
Clean up the default drawing a bit
Quote URL in background-image
Small stylus tweaks
Get closer to PSD's for nodes.

### DIFF
--- a/src/code/utils/js-plumb-diagram-toolkit.coffee
+++ b/src/code/utils/js-plumb-diagram-toolkit.coffee
@@ -3,15 +3,16 @@
 module.exports = class DiagramToolkit
 
   constructor: (@domContext, @options = {}) ->
-    @type      = 'jsPlumbWrappingDiagramToolkit'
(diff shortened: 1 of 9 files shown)


knowuh:
@dougmartin Please have a look.  

The bulf of this is CSS changes and SVG files.

src/code/utils/js-plumb-diagram-toolkit.coffee
```diff
@@ -54,7 +79,7 @@ module.exports = class DiagramToolkit
       @kit.reset()
       @registerListeners()
     else
-      console.log 'No kit defined'
+      console.log "No kit defined"
```
dougmartin: Should this be converted to use log()?

dougmartin:
Looks good other than converting console.log() to log().

knowuh:
Nice catch on the log().

PR #41, opened 2015-05-12
Changed metadata viewer to match comps and fixed other minor bugs

dougmartin:
This is a bit of a mishmash of changes.  The main purpose is to change the image metadata editor to include a dropdown for the license type.  I also replaced the egg, chicken and chick images with public domain images, fixed a "squashed" image in the palette info view that was missed and fixed the hostname lookup in the metadata view.


dougmartin:
@knowuh Please review.  Thanks!

knowuh:
Looks good.

PR #43, opened 2015-05-13
In-place edit  for node titles. [#93909154]

knowuh:
One thing which will need to described and built up is the idea of nodeViewStates.

These will be transient states related to selection for the Nodes in the link-view.

Our current model of selection in link-view is week.  This PR includes some an initial (inadequate) attempt to address some of this.

This PR is for https://www.pivotaltracker.com/story/show/93909154: "When a component is dragged from the palette to the model, it should have a name of 'untitled' and the title should immediately be editable with the cursor focus in the title."

### DIFF
--- a/src/code/views/node-view.coffee
+++ b/src/code/views/node-view.coffee
@@ -1,8 +1,71 @@
-{div, i, img} = React.DOM
+{input, div, i, img} = React.DOM
+tr = require "../utils/translate"
+
+NodeTitle = React.createFactory React.createClass
+  displayName: "NodeTitle"
+
+  getDefaultProps: ->
+    isEditing: false
+    defaultValue: tr "~NODE.UNTITLED"
+
+  getInitialState: ->
+    isEditing: @props.editing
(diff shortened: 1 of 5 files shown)


knowuh:
attn: @dougmartin

src/code/views/node-view.coffee
```diff
@@ -1,8 +1,70 @@
...
+
+  inputElm: ->
+    $(@refs.input.getDOMNode())
+
+  inputValue: ->
+    @inputElm().val()
+
+  updateTitle: (e) ->
+    newTitle = @inputValue()
+    newTitle = if newTitle.length > 0 then newTitle else @props.defaultValue
```
dougmartin: CoffeeScript has one nice conditional assignment sugar you could use here:

    newTitle ||= @props.defaultValue

that evaluates to
   
   newTitle || (newTitle = this.props.defaultValue);

dougmartin:
Looks good.

PR #45, opened 2015-05-14
[#93911110] unify direct & inspector editing for node title.

knowuh:
The inspector title behaved differently form the direct node title editing.
This unifies the string truncation and other behaviors in both contexts.

https://www.pivotaltracker.com/story/show/93911110


knowuh:
@dougmartin -- you are probably swamped, but if you get a second, this is a small one.

dougmartin:
Looks good.  I really like React mixins for both sharing code and isolating view logic.

PR #49, opened 2015-05-21
Selection refactoring initiated by [#94529898]

knowuh:
https://www.pivotaltracker.com/story/show/94529898

New selection-manager and tests.


knowuh:
@dougmartin -- realize you may be busy, in which case I can just ask @jsandoe to have a look.  This is an big commit. There are some whitespace only changes which make it confusing. 

The goal was to begin to break-apart the link-manager. 

I pulled out a selection manager, and added tests for it.

PR #50, opened 2015-05-27
Building Models Tool integration with CODAP

jsandoe:
This branch permits the Building Models Tool to be run within the CODAP environment. The integration does the following:

1. Establishes an IFramePhone connection to codap. If not in CODAP, this IFramePhone connection will simply quietly remain unanswered.
2. Initializes itself with CODAP. Normally this would set up data collections, but we are not ready for that yet, so it basically just establishes initial sizing of the container.
3. Permits saving and restoring of models in CODAP documents.

It does not yet support user level logging. As far as I can tell the existing logging is all developer logging written to the console, so, there wasn't a natural test point. 

The class that implements connection to codap is in the file codap-connect.coffee. I placed it in the 'utils' directory. I placed it there because I could not find a more appropriate place. In the MVC pattern, the code would normally be considered controller logic, so it doesn't fit in the 'views' or 'models' directory. It seemed like the code with a similar role is mainly found in the 'mixins' directory, but this didn't seem right either. Mixins are used in view classes, and so, the separation of concerns (and separation of namespace) is only partial. Interactions with CODAP are more likely to be with the model than the view. Please advise where is the appropriate place for this class.

I construct the CodapConnect singleton in app.coffee. Again, this was motivated by not knowing a better place to put it. A consequence of this is that I could not figure out how to get access to the current pallet since it is, as far as I can tell accessible only through the app-view's status. Please advise on a better place to construct the class.

I had some difficulty integrating IframePhone. It is currently built using browserify, I think, which leaves the require statements in. This causes gulp to report errors and refuse to load changes. I ended up incorporating IframePhone's constituent files. I am not happy with that solution. It would be better to pull in IFramePhone as a submodule, and better yet if its build version was compatible with gulp. Do you know how this is handled elsewhere?

Finally, the current capabilities do not require other objects have access to the CodapConnect singleton, but, presumably, further work in this area will require that the functions of the object are accessible. It was not clear the best way to make this accessible.


src/code/utils/codap-connect.coffee
```diff
@@ -0,0 +1,49 @@
...
+      action: 'initGame'
+      args:
+        name: @name
+        dimensions:
+          width: 800
+          height: 600
+        contextType: 'DG.DataContext'
+    }, @initGameHandler)
+
+  codapRequestHandler: (iCmd, iCallback) =>
```
knowuh: Are most codap interactions handled this way, where Codap initiates the calls? Are there ways for Building models to make calls into Codap?

knowuh:
@jsandoe this looks pretty straight forward.  You are correct that we currently don't have a 'controllers' folder – some other controller like objects live in the models directory right now. (link-manager, selection-manager).  And the undo-redo manager is in Utils.  Some unification here would be good, and maybe a top-level application controller which to provide a single binding interface to the core Building Models functions which are simply delegated to the sub-controllers (link-manager, selection-manager, redo-undo manager, log-manager &etc.)

As far as including iFramePhone, I agree.  There should be a more simple means of including it, and I am not certain what other projects do. It is probably worth discussing on #dev.

I am curious if Codap clients can initiate a call into CODAP. I see the `codapRequestHandler` method, but I was wondering if there was an equivilent `makeCodapRequest` method for clients to use (for example to push log data, or to push state data).

A common idiom in coffeescript is to initialize data in the method named `constructor`. I saw that you had an initialize method. I am wondering if maybe we could do that in the `constructor`.

Thanks for this, I will probably looking closer at this next week.

src/code/utils/codap-connect.coffee
jsandoe: No, the [current API](https://github.com/concord-consortium/codap/wiki/Data-Interactive-API) is mostly initiated from the Data Interactive. Only the save and restore requests come from the CODAP side. 

Some existing Data Interactives implement helper functions to wrap the calls to codap, others build directly on the IFramePhone API. I am not sure what is appropriate here.

jsandoe:
@knowuh, I will move the class to the model directory for now and refactor the initialization to take place in the constructor. Thanks for the helpful comments.

jsandoe:
@knowuh, I moved codap-connect.coffee to the src/model directory and changed it to be initialized in the constructor. I also added a method, 'request', which adds the ability to issue requests to CODAP. The method provides for handling asynchronous responses either by callbacks or Promises, similarly to the Google api.

jsandoe:
@knowuh I fixed a bug I found in this branch. IFramePhone needs to be invoked only after the iFrame has been rendered into the DOM.

knowuh:
@jsandoe I just added iframe-phone using NPM.

PR #52, opened 2015-06-10
First pass on the Node-Value Inspector [#95934234]

knowuh:
This is a First pass on the Node-Value inspector request.  The Node-Value inspector will display the initial value for a node when the simulation is started.  The Node-Value inspector allows students to set the initial value using a slider or text field.

Full details in the PT story:
  https://www.pivotaltracker.com/story/show/95934234

Followup commits will connect nodes to these values.  I am putting in a PR now so that it doesn't get too long.  A similar PR is in the works for the relationship inspector.

@sfentress could you take a look?

### DIFF
--- a/src/code/views/node-value-inspector-view.coffee
+++ b/src/code/views/node-value-inspector-view.coffee
@@ -1,0 +1,71 @@
+{div, h2, label, span, input, p, i} = React.DOM
+
+tr = require "../utils/translate"
+
+module.exports = React.createClass
+
+  displayName: 'NodeValueInspectorView'
+
+  propTypes:
+    max: React.PropTypes.number
+    min: React.PropTypes.number
+    onChange: React.PropTypes.func
+
+
+  getInitialState: ->
+    value: 50
+
+  getDefaultProps: ->
+    max: 100
+    min: 0
+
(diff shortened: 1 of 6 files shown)


knowuh:
Maybe @scytacki you could take a look at this?  I will have aaron or take a look my other WIP branch.

dougmartin:
@knowuh @scytacki - I can look at it also this afternoon if needed.

scytacki:
Doug that would great, thanks.

--
from my phone

src/code/views/node-value-inspector-view.coffee
```diff
@@ -0,0 +1,71 @@
...
+
+  render: ->
+    (div {className: 'value-inspector'},
+      (div {className: 'inspector-content group'},
+        (span {className: 'full'},
+          (label {className: 'right'}, tr "~NODE-VALUE-EDIT.INITIAL-VALUE")
+          (input {
+            className: 'left'
+            type: "number",
+            min: "#{@props.min}",
```
dougmartin: I think you can drop the quotes here.

dougmartin:
Looks good :+1:

PR #53, opened 2015-06-12
Relations inspector

knowuh:
This WIP branch is trying to deliver the 'relationship inspector' as specified here:
https://www.pivotaltracker.com/story/show/95934794

There is also (necessarily) some overlap with this ongoing story: "Calculate model values" for a run:
https://www.pivotaltracker.com/story/show/95934926


knowuh:
Hey @dougmartin  I am sorry to be bombarding you with PR reviews.  Would you mind taking a look at this one too?  I can ask Aaron or Sam to check out the next one; Numerical modeling #54

dougmartin:
I'll look at it this weekend.

src/code/models/relation-factory.coffee
```diff
@@ -0,0 +1,69 @@
...
+    postfixIco: "more-and-more"
+    formulaFrag: "in ^ 2"
+
+  @lessAndLess:
+    id: 4
+    text: tr "~NODE-RELATION-EDIT.LESS_AND_LESS"
+    postfixIco: "less-and-less"
+    formulaFrag: "sqrt(in)"
+
+  @inconName: (incdec,amount)->
```
dougmartin: Should this be iconName instead on inconName?

dougmartin:
Code looks good (I did find a mispelling that I added a line comment about).  I did find some errors while testing:

1. With a line selected if you select any of the right hand tools you get the following console error: "Uncaught TypeError: Cannot read property '_currentElement' of null"
2. With a line selected if you click on background to deselect it you get the same error as above.
3. Once the error above shows none of the tools show after you select a node and then click on the tool.  The same error shows in the console.
4. If you choose "less and less" in the relationship inspector the graph disappears.

knowuh:
Thanks for finding those issues @dougmartin -  I am going to hopefully get them taken care of this afternoon.

knowuh:
@dougmartin -- I ended up taking care of those a bit earlier in my day. af6725e has the fixes.  Thanks again.

PR #54, opened 2015-06-12
Numerical modeling

knowuh:
This PR is for the PT Story: [Evaluate model values #95934926](https://www.pivotaltracker.com/story/show/95934926)

The commits in here are mostly for the Simulation class, and its tests.

An instance of the Simulation class is created with simulation options for duration, time step, and nodes.
This creates a short-lived object created once for each run.

Following a DCI pattern, the Simulation modifies the participating nodes,
and binds them to an integration function, and initializes there state.  We may support
different integration methods in the future. The Current implementation is a simple forward Euler, which seems to work quite well in the tests.

Simulation is the only context in where nodes to have state history, and an integration function. This Simulation class is an attempt to keep all the simulation code close together so its easy to reason about.

It will be easier to review these changes after this upstream changes is merged:

Commits before c2166ba are in relation inspector.
* PR https://github.com/concord-consortium/building-models/pull/52 -- up to  c2166ba  (relation inspector)

### DIFF
--- a/src/code/models/node.coffee
+++ b/src/code/models/node.coffee
@@ -10,7 +10,7 @@ module.exports = class Node extends GraphPrimitive
     if key
       @key = key
     @links = []
-    {@x, @y, @title, @image, @initialValue} = nodeSpec
+    {@x, @y, @title, @image, @initialValue, @color} = nodeSpec
     @color ?= Colors[0].value
 
   type: 'Node'
--- a/src/code/models/node.coffee
+++ b/src/code/models/node.coffee
@@ -5,23 +5,21 @@ tr = require '../utils/translate'
 
 module.exports = class Node extends GraphPrimitive
 
-  constructor: (nodeSpec={x:0,y:0,title:"untitled",image:null}, key) ->
+  constructor: (nodeSpec={x:0,y:0,title:"untitled",image:null,initialValue:50}, key) ->
     super()
     if key
       @key = key
     @links = []
-    @x = nodeSpec.x
-    @y = nodeSpec.y
-    @title = nodeSpec.title
-    @image = nodeSpec.image
-    @color = nodeSpec.color or Colors[0].value
+    {@x, @y, @title, @image, @initialValue} = nodeSpec
+    @color ?= Colors[0].value
 
   type: 'Node'
   addLink: (link) ->
     if link.sourceNode is @ or link.targetNode is @
       if _.contains @links, link
         throw new Error "Duplicate link for Node:#{@.id}"
-      @links.push link
+      else
+        @links.push link
     else
       throw new Error "Bad link for Node:#{@.id}"
 
--- a/src/code/models/simulation.coffee
+++ b/src/code/models/simulation.coffee
@@ -18,11 +18,17 @@ IntegrationFunction = (t) ->
 
 module.exports = class Simulation
 
+  @defaultInitialValue = 50
+
+  @defaultReportFunc = (report) ->
+    log.info report
+
+
   constructor: (@opts={}) ->
     @nodes       = @opts.nodes      or []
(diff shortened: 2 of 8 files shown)


ddamelin:
Euler is fine for many systems models. However, if we want to model physical systems like springs and gravity, then we'll need to use a more accurate integration algorithm. I think the one other's use most as an alternative is RK4: https://en.wikipedia.org/wiki/Runge–Kutta_methods

Typically the integration method is a choice: Euler or RK4. Euler is faster and RK4 is more accurate. I was hoping to perhaps just use RK4 with the assumption that the types of models people might be making with this classroom oriented tool would not be too slow using that method. Given our focus on getting some values generated as quickly as possible, let's stick with this algorithm unless it is easy to switch to RK4.

I was thinking of trying to use the model with CODAP partners and a physical system on June 18, but I was not counting on that, so it would be OK to stick with Euler for now if not easy to update to RK4.

knowuh:
@psndcsrv do you have a moment to look at this?

src/code/models/node.coffee
```diff
@@ -5,23 +5,21 @@ tr = require '../utils/translate'
...
     super()
     if key
       @key = key
     @links = []
-    @x = nodeSpec.x
-    @y = nodeSpec.y
-    @title = nodeSpec.title
-    @image = nodeSpec.image
-    @color = nodeSpec.color or Colors[0].value
+    {@x, @y, @title, @image, @initialValue} = nodeSpec
```
psndcsrv: Looks like this is missing the assignment to `@color`?
`{@x, @y, @title, @image, @color, @initialValue} = nodeSpec`

Or perhaps the line below needs to change back to `@color = nodeSpec.color or Colors[0].value`, or if color is never going to be specified in nodeSpec, then just `@color = Colors[0].value`.

src/code/models/simulation.coffee
```diff
@@ -0,0 +1,93 @@
...
+        duration: @duration
+        timeStep: @timeStep
+        nodeCount: @nodes.length
+      frames: @reportFrames
+      endState: _.map @nodes, (node) ->
+        title: node.title
+        value: node.currentValue
+        initialValue: node.initialValue
+    @reportFunc(data)
+    # TODO CLEANUP
```
psndcsrv: Everything looks pretty clean, so this is probably an unnecessary comment?

test/simulation-test.coffee
```diff
@@ -0,0 +1,251 @@
...
+        describe "nodeB", ->
+          it "should be get A's previous value", ->
+            @nodeB.currentValue.should.equal @nodeA.initialValue
+
+        describe "nodeC", ->
+          it "should get A's original value", ->
+            @nodeC.currentValue.should.equal @nodeA.initialValue
+
+  describe "report", ->
+    describe "for a simple graph A(10) -0.1-> B(0) for 10 itterations", ->
```
psndcsrv: itterations -> iterations

test/simulation-test.coffee
```diff
@@ -0,0 +1,251 @@
...
+          @arguments.duration = 1
+          @simulation = new Simulation(@arguments)
+          @simulation.run()
+
+        describe "nodeA", ->
+          it "should be unaffected", ->
+            @nodeA.currentValue.should.equal @nodeA.initialValue
+
+        describe "nodeB", ->
+          it "should be get A's previous value", ->
```
psndcsrv: should be get -> should get

Also on line 190

psndcsrv:
I think the code looks fine. Most of my comments were just inconsequential spelling/grammar things. I can't really weigh in on the integration methods conversation -- that's not my area of expertise! :) But, it does look like it's doing what you're intending for it to do.

PR #56, opened 2015-06-17
Initial effort to output simulation to CODAP for graphing

knowuh:


knowuh:
@jsandoe  Could you have a look at this, starting with commit ecd572c (second from last)

The last two commits in this history relate to the CODAP integration.  Previous work is related to bare bones simulation running.

PR #58, opened 2015-07-01
Codap undo redo

sfentress:
This PR uses a new CODAP API to allow CODAP to control the app's undo/redo history.

Each time we perform an undoable event, we notify CODAP of this. CODAP can then send us
undo and redo commands, which we perform using our own undo-redo manager.

Our undoable events are stored in CODAP's undo stack, which closely mimics this undo stack, so when CODAP calls 'undo', it pops off the action from CODAP's stack, and causes BuildingModels to pop off the action from its stack. The two will track identically.

We also add the ability for CODAP to inform BuildingModels that it is using its own undo system (which it doesn't always), so that BuildingModels can hide its undo UI. (Note that this doesn't affect whether we send the undoableActionPerformed messages, since these are side-effect free if CODAP isn't using undo.)

We could optionally add to this system a way of passing IDs of events between BuildingModels and CODAP. This would be useful if BuildingModels failed to undo an action, and needed to be aware of that when CODAP tried to undo a second time. 

However, currently BuildingModels does not have any way or expectation that it will "fail" to undo or redo an action. It assumes that an undo or redo will always work perfectly. Therefore, there doesn't seem to be any way that the two systems could get out of sync -- or, at least, that BuildingModels would be aware that it had gotten out of sync. At the moment, therefore, an ID system seems to me to be unnecessary.

### DIFF
--- a/src/code/mixins/app-view.coffee
+++ b/src/code/mixins/app-view.coffee
@@ -98,9 +98,10 @@ module.exports =
       else
         undo = redo = false
       if undo or redo
-        e.preventDefault()
-        @props.linkManager.redo() if redo
-        @props.linkManager.undo() if undo
+        if (@props.linkManager.undoRedoIsVisible)
+          e.preventDefault()
+          @props.linkManager.redo() if redo
+          @props.linkManager.undo() if undo
 
   componentDidUnmount: ->
     @addDeleteKeyHandler false
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -14,7 +14,7 @@ window.initApp = (wireframes=false) ->
     linkManager: LinkManager.instance 'building-models'
     data: getParameterByName 'data'
 
-  opts.codapConnect = new CodapConnect opts.linkManager
+  opts.codapConnect = CodapConnect.instance 'building-models'
 
   appView = AppView opts
   elem = '#app'
--- a/src/code/models/codap-connect.coffee
+++ b/src/code/models/codap-connect.coffee
@@ -121,6 +121,10 @@ module.exports = class CodapConnect
         callback
           success: true
 
+      when 'externalUndoAvailable'
+        log.info 'Received externalUndoAvailable request from CODAP.'
+        @linkManager.hideUndoRedo(true)
+
       when 'undoAction'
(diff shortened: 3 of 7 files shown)


src/code/mixins/app-view.coffee
```diff
@@ -98,9 +98,10 @@ module.exports =
       else
         undo = redo = false
       if undo or redo
-        e.preventDefault()
-        @props.linkManager.redo() if redo
-        @props.linkManager.undo() if undo
+        if (@props.linkManager.undoRedoIsVisible)
+          e.preventDefault()
```
knowuh: If the user presses command-z or ctr-z, does it get handled by CODAP here? 
It would be good to notify CODAP the users intention of wanting to undo the last command if we can.

sfentress: Yes, it gets handled by CODAP.

If CODAP declares that it is using undo/redo, this cases "undoRedoIsVisible" to be false.

This has the effect of not only hiding the buttons, but also causing the button presses to be ignored. (Although actually that would happen anyway, because CODAP catches them, but better to be explicit).

It may be confusing that the property "undoRedoIsVisible" also has an effect on the keyboard presses, so we could change the name if so.

sfentress:
Note: the tests are failing now because the undo-redo manager now calls CODAP-connect on `createAndExecuteCommand`. CODAP-connect has a dependency on iFramePhone, which refers to `window`, which is not in the mocha environment.

This wasn't exposed before because CODAP-connect wasn't ever required in any test.

I'm looking into how best to mock window right now in the tests, but it's a little confusing.

sfentress:
@knowuh The tests are fixed. (Had to mock the `window` object, it's possible that this will have to be watched out for in future tests.)

The undo/redo connection with CODAP is now more robust due to the app being able to inform CODAP if it fails to undo or redo. I had mistakenly thought before that there was no concept of "failing" to undo/redo in Building Models, but actually I see that it's built right in. The result of the action will be passed back to CODAP. If it fails, CODAP will inform the user and clear its own history stacks.

knowuh:
Thanks @sfentress

PR #59, opened 2015-07-22
Save pallet on doc save: [#99430432]

knowuh:
Make the save path of CODAP `saveState` to closely match the path used by the google docs saving.  In order to do this, the Palette management had to be move outside the Application view, and into a new controller.  This controller is using reflux actions and stores.


src/code/models/palette-manager.coffee
```diff
@@ -0,0 +1,76 @@
+LinkManager    = require './link-manager'
+
+# TODO: Maybe loadData goes into some other action-set
+paletteActions = Reflux.createActions(["addToPalette","loadData"])
+
+paletteStore   = Reflux.createStore
+  # NOTE: This Reflux shortcut does not work as advertised,
+  # listenables: paletteActions
```
dougmartin: Did you try "listenables: [paletteActions]"?

knowuh: I did, and I was having no luck with those, see related comments in `palette-manager.coffee`

knowuh:
@dougmartin you had a look at this, are you sort of comfortable with it?

PR #60, opened 2015-07-23
Remove ability for link to self [#98570842]

knowuh:
https://www.pivotaltracker.com/story/show/98570842


knowuh:
@dougmartin super tiny.

PR #61, opened 2015-07-23
Initial Diagram and and Palette should be blank by default. #99715626

knowuh:
https://www.pivotaltracker.com/story/show/99715626


dougmartin:
Looks good :+1:

PR #62, opened 2015-07-23
Allow simple (design only) mode with query parameter #simplified

knowuh:
https://www.pivotaltracker.com/story/show/99430744

User can set design only mode using a query parameter.


dougmartin:
Looks good. :+1:

PR #63, opened 2015-07-29
Move palette to top.

knowuh:
This probably should have gone in as a lot of smaller PRs -- But the work ultimately was fairly atomic, in part because of the refactoring that went into using the PaletteManager.

This is going to require at least one more round of refactoring to clean up the image meta-data caching. The responsibilities for maintaining that cache are now in the Palette manager, but it feels mixed up to me still, especially when serializing the the diagram.

Also, I still haven't been able to get the reflux magic mixins to work as expected.  That will be the very next item I look at.


src/code/models/link-manager.coffee
```diff
@@ -333,9 +328,12 @@ module.exports   = class LinkManager
     linkExports = for key,link of @linkKeys
       link.toExport()
     imageMetadata = {}
-    _.forEach palette, (node) =>
-      if @imageMetadataCache[node.image]
-        imageMetadata[node.image] = @imageMetadataCache[node.image]
+    # The pallete manager is curently in charge of trackign image meta-data
+    PaletteManager   = require "../models/palette-manager"
```
dougmartin: Stylwise shouldn't this require be placed at the top of the page?

src/code/views/node-well-view.coffee
```diff
@@ -7,39 +7,38 @@ module.exports = React.createClass
...
-    topNodePaletteClass    = 'top-node-palette'
+  render: ->
+    topNodePaletteClass    = 'top-node-palette-wrapper'
     topNodeTabPaletteClass = 'top-node-palette-tab'
     if @state.collapsed
-      topNodePaletteClass    = 'top-node-palette collapsed'
+      topNodePaletteClass    = 'top-node-palette-wrapper collapsed'
       topNodeTabPaletteClass = 'top-node-palette-tab collapsed'
-    (div {className: 'top-node-palette-wrapper'},
+    (div {className: 'wrapperwrapper'},
```
dougmartin: Q: What do you call a modern stuttering poet with rhotacism? A: wrapperwrapper

src/code/models/link-manager.coffee
knowuh: I should address this now, its an ugly hack.  Link-manager was requiring palette-manager -- and palette-manager was requiring the link-manager.  This defers the require until its needed.  I don't like it. Circular are a code smell to me. 

We only need this because the link-manager is serializing everything. This is actually why the palette manager required the link-manager in the first place.  

I think the right thing to do would be to create a special controller for Loading & Saving.  I think I can do that as part of this work here:  https://www.pivotaltracker.com/story/show/100146458

dougmartin:
Looks good :+1:

PR #66, opened 2015-08-03
Drag add image

knowuh:
The image browsing dialog now has to support two modes.

In the first mode we are just adding images to the Palette. Nothing is being dragged on the canvas. Its current functionality is great -- we just keep the dialog open so that the user can add more and more images.

The second case is more modal. According to this story: https://www.pivotaltracker.com/story/show/95933006  we need to use the dialog to pick one single image to be dropped onto the canvas (and also added to palette).  

The ImageManager is an attempt to centralize the logic for configuring / showing / hiding the dialog box.  Its also possible to supply a call-back hook now so that the interrupted operation (dragging) can continue after the image is selected.

### DIFF
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -20,7 +20,7 @@ window.initApp = (wireframes=false) ->
     simplified: getParameterByName 'simplified'
 
   opts.codapConnect = CodapConnect.instance 'building-models'
-
+  opts.tree = (require './models/image-manager').tree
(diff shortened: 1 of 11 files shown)


knowuh:
@dougmartin if you get a few minutes, have a look?  thanks!

dougmartin:
Looks good.  I like the ImageManager mixin.

PR #68, opened 2015-08-10
Cleanup and rename PaletteStore and ImageDialogStore.

knowuh:
The PaletteStore saves metadata about prototype nodes used in the Library and Palette. The ImageDialogStore handles state data for the ImageDialog box.

[#100741074]
https://www.pivotaltracker.com/story/show/100741074


knowuh:
@dougmartin I know you are swamped, but if you have a chance to peep this, I would appreciate it.  Again, sort of a large commit.  A lot of the noise is because I renamed the stores and put them in a ./stores/ directory.  Also added a simple test for a foothold.

PR #70, opened 2015-08-13
Seralize relations

knowuh:
(Best to merge prior commits, because this one depends on prior)

Relationship info (is accumulator, initial value, relationship formula, &etc.) is now saved in the JSON.

### DIFF
--- a/src/code/data/migrations/02_add_relations.coffee
+++ b/src/code/data/migrations/02_add_relations.coffee
@@ -1,0 +1,29 @@
+Relationship = require '../../models/relationship'
+
+migration =
+  version: 1.1
+  description: "Adds initial values and relationships."
+  date: "2015-08-13"
+
+  doUpdate: (data) ->
+    @updateNodes(data)
+    @updateLinks(data)
+    data
+
+  # Add initialValue if it doesn't exist
+  updateNodes: (data) ->
+    for node in data.nodes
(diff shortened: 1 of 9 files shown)


knowuh:
@dougmartin @scytacki or @pjanik   have a look?

sfentress:
@knowuh This all looks good to me.

Slightly separate from the PR, the connection between how a formula is specified (`@defaultFormula:    "1 * in"`) and how it's interpreted by the relationship factory (if I understand right, it basically turns that formula into the relationship "increases by about the same") seems a little opaque. I would have thought from seeing the formula `"1 * in"` that I could equally specify `"2 * in"`, but if I understand right, that wouldn't be acceptable by the relationship factory. If we're limited to a static set of formula components, making a formula explicitly use those components, with like `"INCREASES BY_THE_SAME`" might make that clearer.

On the other hand, if this is all under the hood maybe it makes no difference, and I could also not have a complete picture of the formulas and relationships yet.

knowuh:
@sfentress  -- Thanks.  The Formula & Relationship work isn't complete. I like the idea of using named constants as an improvement.   We have to support visually constructing formulas, and converting formulas back into simple verbal & pectoral statements.  Sounds like you are going to be headed into that code soon enough.  Improvements throughout are welcome.

PR #71, opened 2015-08-18
Delete palette items

knowuh:
I am putting this PT in now for [#100142746] -- "A user should be able to delete items from the palette."

It doesn't handle re-assigning the existing nodes the replacement image.  I am working on that now.

see: https://www.pivotaltracker.com/story/show/100142746


knowuh:
@sam -- Can you please give this a read?

These changes support this story: "A user should be able to delete items from the palette." https://www.pivotaltracker.com/story/show/100142746

They add a new store, which manages the palette-delete-view.

The link manager listens for changes in this store, and if a palette item has been deleted and a replacement image chosen, it updates matching nodes in the graph to use the replacement paletteItem.

PR #72, opened 2015-08-21
Delete palette items2

knowuh:
Fixes for Delete-palette-items [#100142746]:

- Add undo / redo suport.
- Context sensitive wording and icons for replace / delete (paletteItem)

### DIFF
--- a/src/code/models/link-manager.coffee
+++ b/src/code/models/link-manager.coffee
@@ -33,7 +33,7 @@ module.exports   = class LinkManager
 
   paletteDelete: (status) ->
     {deleted,paletteItem,replacement} = status
-    if deleted and paletteItem
+    if deleted and paletteItem and replacement
       for node in @getNodes()
         if node.paletteItemIs paletteItem
           @changeNode({image: replacement.image},node)
--- a/src/code/models/link-manager.coffee
+++ b/src/code/models/link-manager.coffee
@@ -169,7 +169,7 @@ module.exports   = class LinkManager
 
   _removeNode: (node) ->
     delete @nodeKeys[node.key]
-    NodesStore.actions.nodesChanged()
+    NodesStore.actions.nodesChanged(@getNodes())
     for listener in @nodeListeners
       log.info("notifying of deleted Node")
       listener.handleNodeRm(node)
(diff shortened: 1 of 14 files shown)


knowuh:
This PR introduces a nodes-store --  Right now it only tracks which nodes are on the canvas, and if they are of the same 'paletteItem' (type) as the one selected in the palette / meta-data view.

This store was needed so that the UI elements related to 'delete or replace' of PaletteItems could check if the current PaletteItem was in use in the Graph.

It would be great if the node-store's responsibility could grow to include some of the work being done in the old link-manager. (Actions related to adding and removing nodes for example). With a little more work I am sure it could.

@sfentress if you have a chance for a quick look, I would appreciate it.

knowuh:
Still related to [#100142746](https://www.pivotaltracker.com/story/show/100142746)

sfentress:
@knowuh This all looks fine to me.

From a user perspective, the only odd thing is that the undo/redo actions operate on the under-the-hood actions, not what the user actually did:

1. You have two nodes of type A
2. You perform one action, Replace Type A with Type B
3. This results in three actions that have to be undone individually:
  1. First the original image is re-added to the palette
  2. Then one node is reverted
  3. Then the second node is reverted

I don't know whether or not it's related to this story, but perhaps we should add the concept of a set of actions that can be added to the undo history, which, even if they are stored as n-different actions under the hood, are all undone or redone with a single click of undo/redo?

sfentress:
@knowuh Also, a small bug:

1. Add a node of type A
2. Replace node with type B
3. Undo (twice) so that the node of type A is visible on the canvas
4. From the palette menu, A can only be deleted, not replaced, even though a node of type A is on the canvas. (The resulting action actually does replace, but this is not evident to the user, and the user cannot select the replacement.)

knowuh:
Thanks @sfentress -- Dan had a few comments too, so expect some updates later.

knowuh:
> I don't know whether or not it's related to this story, but perhaps we should add the concept of a set of actions that can be added to the undo history, which, even if they are stored as n-different actions under the hood, are all undone or redone with a single click of undo/redo?

Yeah, that has been missing for a while.  Not sure I want to tackle it as part of this story or not.

sfentress:
> Not sure I want to tackle it as part of this story or not.

My vote would certainly be that that's a separate feature for separate prioritization.

PR #73, opened 2015-08-25
Accumulators sum their inputs independently; +tests

sfentress:


test/simulation-test.coffee
```diff
@@ -64,11 +64,60 @@ describe "Simulation", ->
       describe "the result", ->
         it "should give B 10 at the end", ->
```
knowuh: Forgot to update the above text I think?

knowuh:
Looks good.  Line 65 of simulation-test.coffee should probably be rewritten to read `it "should give B 1 at the end", ->`  when you get around to it.  But I am merging as is.

PR #74, opened 2015-08-25
Show Run button only in codap

sfentress:
We now only show the Run button if we are embedded in CODAP. We do this by listening to a successful phone call to CODAP, and setting state via a new codap-store.

Also refactored how we respond to CODAP asking us to hide the undo/redo buttons into the codap-store.

https://www.pivotaltracker.com/story/show/101447232


knowuh:
Looks good :+1:

PR #75, opened 2015-08-26
LinkManager → DiagramStore refactor

knowuh:
These are a lot of small changes, mostly renaming things, but also moving over to the reflux store pattern for the DiagramStore (LinkManager).

@sfentress I hope this doesn't make you blind.


sfentress:
Outside possible further discussion on that comment above, this all looks good to me. I'm going to pull this in now, so future work today doesn't require complex rebasing.

PR #76, opened 2015-08-26
Batch undo redo

sfentress:
Allows us to batch multiple undoable commands into a single command that can be undone or redone in one click.

https://www.pivotaltracker.com/story/show/101973812

@knowuh Let me know if you have a better solution for the `setTimeout` that is called in `endCommandBatch`. Without it, if we call

    @undoManger.startCommandBatch()
    PaletteStore.actions.deleteSelected()
    ... anything else
    @undoManger.endCommandBatch()

then `deleteSelected` gets called after `endCommandBatch`, because it uses an async callback.

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -194,6 +194,12 @@ GraphStore  = Reflux.createStore
     if node
       @changeNode(data,node)
 
+  startNodeEdit: ->
+    @undoRedoManager.startCommandBatch()
+
+  endNodeEdit: ->
+    @undoRedoManager.endCommandBatch()
+
   selectLink: (link) ->
     @selectionManager.selectLink(link)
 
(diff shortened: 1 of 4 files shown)


sfentress:
Actually, I realize one solution could be to make undoRedo a store, and call `endCommandBatch` via an action, which I assume would push it to the back of all the other actions queued up.

It would be a bit odd, because we'd need `startCommandBatch` to be called instantly, so that can't be done as an action. So you'd have something like

    undoManger.startCommandBatch()
    ... actions
    undoManger.actions.endCommandBatch()

But maybe this would still be cleaner? To me it looks uglier and more confusing, rather than just relying on endCommandBatch just working without caring about internals, but not sure.

knowuh:
This looks very good. I think we should merge it as is.

The setTimeout is well documented, and doesn't bother me really.  

Eventually moving to an 'all reflux actions' based approach seems like a good idea. If all our undoable actions take place reflux actions, then maybe `startCommandBatch` could be implemented as an action too?  

But I don't like the asymmetry of having `endCommandBatch` being a reflux action, and `startCommandBatch` being simple method.

PR #80, opened 2015-09-04
Min max editing

sfentress:


knowuh:
Looks great.

src/code/views/node-value-inspector-view.coffee
```diff
@@ -33,6 +33,46 @@ module.exports = React.createClass
   selectText: (evt) ->
     evt.target.select()
 
+  renderEditableProperty: (property, classNames) ->
+    swapState = =>
+      @setState "editing-#{property}": not @state["editing-#{property}"], ->
+        focusable = React.findDOMNode(this.refs.focusable)
+        focusable.focus() unless not focusable
```
dougmartin: CoffeScript has a nice shortcut for this:
    focusable?.focus()

sfentress: True, forgot about that! Thanks.

PR #83, opened 2015-09-11
New simulation engine

sfentress:
This PR is for the stories

https://www.pivotaltracker.com/story/show/101699204
https://www.pivotaltracker.com/story/show/101698452
https://www.pivotaltracker.com/story/show/101699240

It first redefines the simulation engine so that dependent variables (nodes with incoming links that are not collectors) always get their values from their parents (recursively if necessary); this then introduces the idea of invalid graphs (graphs with a closed loop of dependent variables), so we need to test this and inform the user if the graph is invalid; and finally we hide the value editing of dependent variables, so that the user doesn't incorrectly think they are making a meaningful change by changing the initial values.

### DIFF
--- a/src/code/models/simulation.coffee
+++ b/src/code/models/simulation.coffee
@@ -97,10 +97,48 @@ module.exports = class Simulation
 
     @reportFunc(data)
 
+  # Tests that the graph contains no loops consisting of dependent variables.
+  # A graph such as A->B<->C is invalid if B and C connect to each other and
+  # neither are accumulators
+  graphIsValid: ->
+    _.each @nodes, (node) -> node._isValid = null
+
+    # Recursive function. We go throw a node's non-independent ancestors, and if
+    # we find ourselves again, or if any of our ancestors find themselves again,
+    # we have a loop.
+    isInALoop = (node) ->
+      linksIn = node.inLinks()
+      # quick exit if we're not a dependent variable, or if we've already been checked
+      return false if node._isValid or node.isAccumulator or linksIn.length is 0
+      for seen in nodesSeenInSegment
+        if seen is node
+          return true
+
+      nodesSeenInSegment.push node
+
+      for link in linksIn
+        return true if isInALoop link.sourceNode
+
+      return false
+
+    for node in @nodes
+      nodesSeenInSegment = []
+      return false if isInALoop node
+      # if node was not in a loop, none of its ancestors were either,
+      # so mark all nodes we've seen as valid for speed
+      for seen in nodesSeenInSegment
+        seen._isValid = true
+
+    return true
+
+
   run: ->
     time = 0
     @reportFrames = []
     _.each @nodes, (node) => @initializeValues node
+    if not @graphIsValid()
+      # We should normally not get here, as callers ought to check graphIsValid themselves first
+      throw new Error("Graph not valid")
     while time < @duration
       _.each @nodes, (node) => @nextStep node  # toggles previous / current val.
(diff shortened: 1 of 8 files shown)


sfentress:
@knowuh Can you review if you get a chance?

src/code/models/simulation.coffee
```diff
@@ -89,10 +97,49 @@ module.exports = class Simulation
...
+
+      nodesSeenInSegment.push node
+
+      for link in linksIn
+        return true if isInALoop link.sourceNode
+
+      return false
+
+    for node in @nodes
+      nodesSeenInSegment = []
```
knowuh: Does this declaration mask the array of the same-name on line 104? I feel like this only effects marking the individual nodes as invalid -- so its probably not critical.

test/simulation-test.coffee
```diff
@@ -66,237 +66,125 @@ describe "Simulation", ->
...
+        ]}
+
+        # three-node graph (>-) with non-linear averaging
+        {A: 10, B: 20, C: null, AC: "1 * in", BC: "0.1 * in",
+        results: [
+            [10, 20, 6]
+            [10, 20, 6]
+        ]}
+
+        # three-node graph with collector (>-[C])
```
knowuh: Just a note: I find it interesting that dependent variables will average their inputs, while an accumulator will sum. It makes sense, but it gave me pause.

test/simulation-test.coffee
```diff
@@ -66,237 +66,125 @@ describe "Simulation", ->
...
+        results: [false]
+        }
+      ]
+
+      _.each scenarios, (scenario, i) ->
+        it "should compute scenario #{i} correctly", ->
+          nodes = {}
+          for key, value of scenario
+            if key.length == 1
+              isAccumulator = typeof value is "string"
```
knowuh: I might not understand isAccumulator.  I would expect to assert that "+" is in the string.

knowuh:
@sfentress this looks Great.  I had a few minor questions.  Feel free to merge it.

test/simulation-test.coffee
sfentress: I think it's quite possible that users will also find this confusing, but I think we'll wait for @ddamelin and others to try it out for a while, and see if they think it still makes sense.

test/simulation-test.coffee
sfentress: Yeah, I ought to be more explicit here. The value of "50+" looked clear to me as a human, to differentiate itself clearly from non-accumulators, but then I figured, since only accumulators get string values, why not just check if it's a string?

I'll add the explicit check for the +, to formalize it.

src/code/models/simulation.coffee
sfentress: Actually, this is the critical declaration, where as the one on line 104 shouldn't be there.

It's on 104 as well because of my occasional uncomfortableness of coffeescript -- not being able to place an explicit `var` declaration yourself. To my mind, it's clearer to see `var nodesSeenInSegment` at the top of this function, because it makes it clear that it's declared outside the scope of the inner function, whereas that's not immediately apparent when the initial declaration is below the inner function.

But I should remove it, because it's non-functional. Coffeescript will correctly put the declaration at the top of the function anyway, so I shouldn't worry about it.

(In terms of it's function, which is also maybe a little confusing: nodesSeenInSegment is all the nodes we have encountered when looking at one node's ancestors. This gets cleared for each new node we ask for, although these subsequent lookups for the other nodes ancestors will generally be much faster because many of them may already have been marked as good.)

PR #84, opened 2015-09-15
Include private URL in addressbar when saving document.

knowuh:
@sfentress  if you could take a look when you get a moment.

This shouldn't have been a big commit, but it ended up being a bit complicated.

This commit ensures that everytime the file is saved, the parameters in the address hash are updated to reflect the new save location. This is true for public and private documents.  It will also remove public urls from the address bar one they have been loaded by an anonymous user, so that they don't have the idea that their changes are also saved.

Additionally this commit makes it so that the "open in codap" button is disabled unless the document is saved.

The largest chunk of new work is:  hash-parameters.coffee  which was written to simplify loading / saving of document properties in the URL in a non-conflicting manner.  hash-parameters.coffee also has some rudimentary tests.


knowuh:
Thank's @sfentress. Good point regarding HashParams.  See 818245b for fix.

sfentress:
All looks good to me. I was thinking maybe `getParams` could check if it already had checked the latest parameter string (just compare it to a lastParams string), but calling `fromLocationHash` takes essentially zero time, so that's probably just an over-complication.

PR #87, opened 2015-09-17
Add serialization to settings

sfentress:


src/code/stores/app-settings-store.coffee
```diff
@@ -3,6 +3,7 @@ AppSettingsActions = Reflux.createActions(
     "showSettingsDialog"
     "close"
     "capNodeValues"
+    "importSettings"
```
knowuh: If we are following the actions pattern (don't feel strongly about this) then the method `importSettings` maybe should be renamed `onImportSettings`.

knowuh:
This looks good to me, but see inline comment about `importSettings` action.

src/code/stores/app-settings-store.coffee
sfentress: Actually it was originally an action, was called as an action, and the method was `onImportSettings`. However, I was tripped up by the async nature of actions during tests.

I actually solved the issue, using Mochas `done` function, which can be asynchronous, but then I felt it was convoluted, and that loading settings should be a synchronous event, like loading nodes and links is.

I switched the importer to call `importSettings` directly, but forgot about the action string above. I will remove it.

sfentress:
Removed the importSettings action string.

PR #89, opened 2015-09-21
Put 'Diagram-only' setting in settings panel [#103626652]

knowuh:
@sfentress please review when you have a moment.

This is simple commit, it moves the 'diagram-only' setting into the settings panel.

https://www.pivotaltracker.com/story/show/103626652

### DIFF
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -21,7 +21,6 @@ window.initApp = (wireframes=false) ->
     publicUrl: HashParams.getParam 'publicUrl'
     data: HashParams.getParam 'data'
     googleDoc: HashParams.getParam 'googleDoc'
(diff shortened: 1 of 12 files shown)


sfentress:
@knowuh This looks fine.

Should the setting be serialized? If someone creates a simplified model, do they expect others to open it as such?

Besides that question, this is all good.

knowuh:
@sfentress -- I also was wondering if we would want to serialize the 'diagramOnly' mode or or not.  
At first it seemed like a simple URL parameter. But in thinking about it, it seems to make sense as part of the model data too.

ddamelin:
Yes we should serialize the advanced settings, including diagram only.

sfentress:
Looks good to me. :+1:

PR #90, opened 2015-09-23
Initial effort on custom slider component.  [#102293636]

knowuh:
@sfentress can you give this a look-over when you get a chance?

Its a custom slider control which closely matches the PSD's from Blender Box.

By browsing to http://localhost:8080/test.html  you should see an interactive version, which has been tested in FF, Safari, and Chrome.

_local__building_models_demo

https://www.pivotaltracker.com/story/show/102293636

### DIFF
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -1,5 +1,6 @@
 AppView     = React.createFactory require './views/app-view'
 
+ValueSlider  =require './views/value-slider-view'
 GraphStore   = require './stores/graph-store'
 CodapConnect = require './models/codap-connect'
 HashParams   = require './utils/hash-parameters'
--- a/src/code/views/value-slider-view.coffee
+++ b/src/code/views/value-slider-view.coffee
@@ -1,6 +1,6 @@
 {div, i, label, span, input, svg, circle, path, rect} = React.DOM
 
-SVGSlider = React.createFactory React.createClass
+ValueSlider = React.createClass
   displayName: 'SVGSlider'
 
   getDefaultProps: ->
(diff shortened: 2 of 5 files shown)


sfentress:
This looks great, and I having a test page. 

I'm slightly confused by it (the test page), though. When we are creating a new component and want to test it, it looks like we need to find the last component that was tested, remove the `window.testComponent` line from there, and add it to out own component?

That's fine, but it's a little unfortunate that we have to find and change an unrelated view. Do you think it would make any sense to have, say, an set of components that you might want to test, and then you can add to that set independently? Something like

    // at the bottom of a component you want to test
    window.testComponents ?= {}
    window.testComponents.ValueSlider = (domID) -> React.render myView({}), domID

    // in test.html
    var testComponent = window.testComponents.ValueSlider;   // change this to test a different view
    testComponent($("#test")[0]);

That seems to work, do you think it makes any sense? Or maybe it's not really an issue at all.

Even better would be if we could just refer to ValueSlider directly in test.html, without having to modify value-slider-view.coffee at all, but I get confused about whether it's possible to do that with `require`.

sfentress:
Another thing that may or may not be under this PR (it could also be saved for some later "make it good for mobile" push) is that the component works fine with mobile, but is definitely harder to consistently initiate a drag, compared to the existing slider component. 

One big difference, I notice, is that the native browser slider component starts a drag when you hit anywhere on the line -- the handle jumps to that point, and remains captured when you move. This works without you realizing it when you miss the handle by a few pixels (and also allows you to change the value without actually dragging, but just clicking). With this component a drag is only initiated when you correctly press and drag the handle.

Again, not sure whether that's part of this specific PR -- it sounds like potentially a lot of extra work... :grimacing:

knowuh:
Thanks for the info about not working optimally on mobile.  I (of course!) want to kick that down the road so that we can get to 'live mode' or whatever it is called sooner rather than later.

As far as the test page, I agree it would be good to have a simpler method for this.  

I think your suggestion is a good one. 

I had the idea for a global array of components that register themselves for testing, and then either iterate through that list, adding each to the dom, or provide a UI for selecting on a component on the test page. 

But actually, now that I am testing another widget, its not that big of an issue to add & remove the code manually for the component under test.

ddamelin:
I worry some about kicking the mobile can down the road as we develop UI components, because the repercussions might be that we have to totally redesign the component and the spacing. It would be OK to leave that for the near future, but not for too long.

sfentress:
I don't see this as threatening any layout changes, it's just a question of how best to detect the touches, so I think improvements to it could be isolated in any future story.

:+1:

PR #91, opened 2015-09-25
Simulation panel

sfentress:
This adds the Simulation Panel UI, as requested by https://www.pivotaltracker.com/story/show/103626652

No new functionality exists in this panel, only the run button works. The other buttons show disabled states, but the sliders don't show a disabled state, even though they are non-functional. If we think this will be confusing, I can add a disabled state to the sliders for the window of time where we have non-useful sliders.

I would normally want to keep working on this branch to add more functionality to the panel, but since it contains features useful to @knowuh, and since @ddamelin has indicated that the panel alone is a useful feature (to allow users to imagine future development) I think this could be ready as is?

### DIFF
--- a/src/code/stores/simulation-store.coffee
+++ b/src/code/stores/simulation-store.coffee
@@ -1,0 +1,39 @@
+SimulationActions = Reflux.createActions(
+  [
+    "expandSimulationPanel"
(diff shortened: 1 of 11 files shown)


knowuh:
Looks good. :+1:

PR #92, opened 2015-09-29
Simulation editable steps

sfentress:
This adds an editable range to the slider, which is then used to set the number of steps the model will run for.

The editable range slider is modeled after the existing one in the node properties panel.

This also includes a slight refactoring of the slider. This isn't a necessary refactoring, but after a fair bit of time delving into best practices for React components, I think it makes sense.

This doesn't quite complete https://www.pivotaltracker.com/story/show/102293096, but it adds the most valuable feature (settable number of steps), and I think it makes sense as its own PR to encapsulate the slider changes.

https://www.pivotaltracker.com/story/show/102293096
https://www.pivotaltracker.com/story/show/102606120

@knowuh When you have a chance, can you take a look?

### DIFF
--- a/src/code/app.coffee
+++ b/src/code/app.coffee
@@ -23,7 +23,6 @@ window.initApp = (wireframes=false) ->
     data: HashParams.getParam 'data'
     googleDoc: HashParams.getParam 'googleDoc'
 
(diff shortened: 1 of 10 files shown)


knowuh:
Looks good.  :+1:

PR #93, opened 2015-09-29
Add sliders to nodes

knowuh:
This work is part of https://www.pivotaltracker.com/story/show/102293636

While not complete, I think it would be worth integrating now.

This PR also improves some of the Node styling, so its closer to what is in the PSD's.

Known deficiencies:

* CSS / Styling issues remain for nodes & sliders in some states.
* The slider values are not connected to simulation properties.

@sfentress  when you get a chance, could you review?


sfentress:
@knowuh Besides the two questions above, this all looks good to me. I really like the new UX of the nodes.

knowuh:
@sfentress  you might just want to look this commit, where I addressed the mixin question: https://github.com/concord-consortium/building-models/commit/2ae7da1bc67ddab5c82ba15b09dcf360e72f52bc

sfentress:
Looks good to me!

PR #94, opened 2015-09-29
Sliders are connected to node initial-values.  [#102293636]

knowuh:
Also part of https://www.pivotaltracker.com/story/show/102293636.  This time we have connected the sliders in the graph window to the nodes initial-values.  

There is more work to be done, but I thought I would put this up to talk about.  At the time of this PR, I also deployed to http://concord-consortium.github.io/building-models/

initial-value-sliders

Changes:  

* Sliders only show up for independent variables, or accumulators.
* Sliders adjust the 'initial value' of nodes.
* The Value text-edit field is the 'current-value' -- which is usually just the initial value.

### DIFF
--- a/src/code/models/node.coffee
+++ b/src/code/models/node.coffee
@@ -7,9 +7,9 @@ module.exports = class Node extends GraphPrimitive
   # Properties that can be changed.
   @fields: [
     'title', 'image', 'color', 'paletteItem',
-    'initialValue', 'min', 'max', 'isAccumulator',
-    'valueDefinedSemiQuantitatively']
-    
+    'initialValue', 'min', 'max', 'value',
+    'isAccumulator', 'valueDefinedSemiQuantitatively']
+
   constructor: (nodeSpec={x:0,y:0,title:"untitled",image:null,initialValue:50,min:0,max:100,isAccumulator:false, valueDefinedSemiQuantitatively:false}, key) ->
     super()
     if key
@@ -91,6 +91,9 @@ module.exports = class Node extends GraphPrimitive
       isAccumulator: @isAccumulator
       valueDefinedSemiQuantitatively: @valueDefinedSemiQuantitatively
     key: @key
+
+  canEditValue: ->
+    @inLinks().length is 0 or @isAccumulator
 
   paletteItemIs: (paletteItem) ->
     paletteItem.uuid is @paletteItem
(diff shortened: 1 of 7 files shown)


sfentress:
This all looks great to me. The only thing that I could imagine being a part of the PR is being able to set that the Slider's values are all integers. (Or even be able to define that it has n-steps.) It's a pretty common slider functionality, and we'll probably need it soon anyway. When setting the values using the slider, it's pretty much impossible to set them to integers, which feels unexpected, especially as the view tells us it's an integer.

image

sfentress:
This all looks great to me. The only thing that I could imagine being a part of the PR is being able to set that the Slider's values are all integers. (Or even be able to define that it has n-steps.) It's a pretty common slider functionality, and we'll probably need it soon anyway. When setting the values using the slider, it's pretty much impossible to set them to integers, which feels unexpected, especially as the view tells us it's an integer.

screenshot 2015-09-29 18 36 27

knowuh:
That is a good point about the sliders.  I like the idea of being able to set the increment of the slider.  I think I will add that to this PR.

knowuh:
@sfentress I added increment values to the sliders in https://github.com/concord-consortium/building-models/commit/d0f1042459f4f3c2d1968df7186b248599fa7d90

I also changed the variable layout again in https://github.com/concord-consortium/building-models/commit/596a91fa5884e3a1e79304670715508164b89a86

Please check it out if you have time.

sfentress:
The slider changes look great.

Is it just me, or do the new arrows look really far away from the nodes? If that's the design, though, then that's totally fine and this all looks good.

Old network:

screenshot 2015-09-30 22 21 32

New network:

screenshot 2015-09-30 22 21 40

It's just confusing because in your top image above, the edge between the egg and chick seems to connect directly to the egg, where as here it's far away, so I don't know if there's different states it can be in that might cause that.

Besides that question, though, :+1:

knowuh:
Yeah, there is some trade-off that we can discuss in the meeting tomorrow WRT where the arrows display.  Basically they connect to the outside edge of a DIV.  In this case I decided to make an invisible div which encompassed the tittle and the image. Previously it was just the image.  There are also different states, so when there are sliders and other things, I change the div that the arrows point at...  TBD, but I am going to merge.

sfentress:
Totally agree.

PR #95, opened 2015-09-30
Simulation time scale

sfentress:
This adds the time units drop downs, and allows a user to run a simulation for, e.g. 2 weeks with step size of 5 hours.

https://www.pivotaltracker.com/story/show/102293096

### DIFF
--- a/src/code/views/dropdown-view.coffee
+++ b/src/code/views/dropdown-view.coffee
@@ -9,7 +9,8 @@ DropdownItem = React.createFactory React.createClass
 
   render: ->
     className = "menuItem #{if @props.isActionMenu and not @props.item.action then 'disabled' else ''}"
-    (li {key: @props.name, className: className, onClick: @clicked }, @props.name)
+    name = @props.item.name or @props.item
+    (li {className: className, onClick: @clicked }, name)
 
 module.exports = DropDown = React.createClass
 
@@ -54,7 +55,7 @@ module.exports = DropDown = React.createClass
       )
       (div {className: menuClass, onMouseLeave: @blur, onMouseEnter: @unblur},
         (ul {},
-          (DropdownItem {name: item.name or item, item: item, select: @select, isActionMenu: @props.isActionMenu}) for item in @props.items
+          (DropdownItem {key: item.name or item, item: item, select: @select, isActionMenu: @props.isActionMenu}) for item in @props.items
         )
       )
     )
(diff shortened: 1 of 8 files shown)


src/code/views/dropdown-view.coffee
```diff
@@ -46,7 +54,41 @@ module.exports = React.createClass
       )
       (div {className: menuClass, onMouseLeave: @blur, onMouseEnter: @unblur},
         (ul {},
-          (DropdownItem {key: item.name, item: item, select: @select}) for item in @props.items
+          (DropdownItem {name: item.name or item, item: item, select: @select, isActionMenu: @props.isActionMenu}) for item in @props.items
```
knowuh: I think we need to maintain `key` as an attribute, because in some cases (when selecting a palette item for a node I think) we dynamically update this list.  I could be wrong, but I have noticed console warnings about this in the past.  See https://facebook.github.io/react/docs/multiple-components.html#dynamic-children

src/code/views/simulation-panel-view.coffee
```diff
@@ -92,11 +100,17 @@ module.exports = React.createClass
...
-            )
-            (Dropdown {anchor: "Year", items: []})
+              style: {width: "#{Math.max 3, (@state.stepSize.toString().length+1)}em"}
+              value: @state.stepSize
+              min: "0"
+              onChange: @setStepSize
+            })
+            (Dropdown {
+              isActionMenu: false
+              onSelect: SimulationStore.actions.setStepUnits
```
knowuh: Curious why one of these used `onSelect` and the other used `onChange`.

knowuh:
This looks great @sfentress .  I haven't yet confirmed my hypothesis about the `key` attribute for selection items.

src/code/views/simulation-panel-view.coffee
sfentress: The diff makes it look like there are two DropDowns defined there, but the `onChange` actually belongs to the input defined above, while `onSelect` is for the DropDown.

Do you think DropDown should call `onChange` instead of `onSelect`? I've been going back and forth whether form-like components should have a call that's semantically meaningful to the type of action, or whether they should all use a generic `onChange`. The ValueSlider, for instance, now accepts both `onValueChange` and `onRangeChange`, but they could both be melded into `onChange`.

src/code/views/dropdown-view.coffee
sfentress: It looks like you're right about that. I changed it to `name` because I wanted to *use* that attribute, but when you set `key` on a component it turns out it's not available at `@props.key`. Instead of figuring out why, I chose to use `name`.

So if I want to pass in a name I'll have to pass *both* key and name. Or, I guess, just re-calculate name in the view.

I will update this now.

knowuh:
Looks great! :+1:

PR #96, opened 2015-09-30
Make undo and redo into actions, fix several bugs

sfentress:
This PR turns undo and redo into Reflux actions, and along the way fixes several bugs:

1. Closing the delete dialog without deleting anything called undo() on the last action
2. Opening the delete dialog, adding a new image, then cancelling and closing the dialog wasn't correctly removing the newly-added image
3. Even when the newly-added image was correctly removed, the use could re-add the item via the "redo" button, which would be unexpected as the user never explicitly undid anything
4. Clicking away from a node without changing its title was adding an empty action to the undo stack. (This was unrelated to the above, but fixed anyway.)

@knowuh When you have a little time, could you hammer on this for a new minutes and see if you catch any edge-cases where I might have added a regression? I haven't noticed any issue caused by turning undo/redo into actions, but I may have missed something. I don't think this is urgent, though.

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -177,9 +177,16 @@ GraphStore  = Reflux.createStore
         isAccumulator: node.isAccumulator
         valueDefinedSemiQuantitatively: node.valueDefinedSemiQuantitatively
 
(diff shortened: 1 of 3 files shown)


knowuh:
This looks good, and I don't see any show stoppers while playing with it locally.  :+1:

PR #97, opened 2015-10-01
Remove font awesome

sfentress:
This should replace all remaining fa icons, and removes the bootstrap font-awesome stylesheets.

@knowuh Can you take a quick glance?

https://www.pivotaltracker.com/story/show/104119738


knowuh:
Looks good & works. :+1:

PR #98, opened 2015-10-02
Fixes bug where sliders may display in 'simple mode' #104658206

knowuh:
https://www.pivotaltracker.com/story/show/104658206

@sfentress this is a trivial change  (99% of it is whitespace)


sfentress:
:+1:

Fully approve of white-space only fixes :)

PR #99, opened 2015-10-07
Reflux actions for loading. Serialize simulation [#104657406]

knowuh:
###  Action / Listener changes:

* Import is now an action.
* Broke out  graph-actions to resolve circular dependencies.
* Instead of listening for onLinksChange or onModelChange, just listen for onGraphChanged.

###  Changes to SimulationStore:

* Moved CODAP initialization into _sendSimulationData, to isolate that dependency.

* Broke out _setUnitsNames method, initially so names could be calculated after loading from serialized form.  Even though its overlaps functionally with `time-units-test.coffee`, I made a separate test for it.

There were also new migrations and serializations added.

See: https://www.pivotaltracker.com/story/show/104657406 for PT story.

### DIFF
--- a/src/code/actions/graph-actions.coffee
+++ b/src/code/actions/graph-actions.coffee
@@ -1,0 +1,1 @@
+module.exports = Reflux.createActions [ "graphChanged" ]
--- a/src/code/actions/import-actions.coffee
+++ b/src/code/actions/import-actions.coffee
@@ -1,0 +1,1 @@
+module.exports = Reflux.createActions [ "import" ]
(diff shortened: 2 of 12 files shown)


knowuh:
@sfentress  – I ended up modifying simulation-store more than I probably should have.  Looks like we were both in there.  If you want to submit a PR first for your work in `simulation-speed`, I can rebase this on-top of your work.

sfentress:
@knowuh Your changes look pretty self-contained. I'm still adding one more thing to my PR, so I'll look at this one first, and then rebase mine on top of yours when I'm done.

sfentress:
I think this all looks great. :+1:

Just to clarify, this doesn't trigger "serialize" as an action that all the stores listen to, right? The actual serialization process is structurally unchanged, right?

PR #100, opened 2015-10-07
Simulation speed controls

sfentress:
This adds the ability to run the model at a slower speed, sending data to CODAP one or more rows at a time.

https://www.pivotaltracker.com/story/show/102294388

@knowuh Could you take a look at this when you have the time?


knowuh:
This looks great. 

I don't think I understand why the the mocha async callbacks like 
`it "should blah blah", (done) ->`  get called more than once, and need to be tracked by the local vars `calledback` -- but I take your word for it.

knowuh:
Looks great!

PR #102, opened 2015-10-09
Unified icons:  Pull icons from `codap-ivy-icons`

knowuh:
This PR: 

* Updates the icon fonts to use the newest set.
* Fixes some alignment issues (with the turtle and rabbit mostly)
* Replaces the old inspector tool icons with the actual blender box icons.
* Includes a shell script & instructions to pull down latest font.

@sfentress if you have a chance to review.  :)


sfentress:
@knowuh This all looks good.

The only thing that looked odd was that the simulate button ended up very small. I pushed a proposed tweak which makes it look closer to BlenderBox's design (https://basecamp.com/2600671/projects/8704697/messages/47136989?enlarge=177280745#attachment_177280745). Does that look ok to you? (Of course, it's basically moot, since we're about the change that, but I figured it made sense to keep it looking the same as before until we make the change.)

Besides that, this is all :+1:

knowuh:
Your proposed tweak looks good to me @sfentress ,  I leave it to you to merge when you want.

PR #105, opened 2015-12-17
Version and build info

knowuh:
@sfentress  If you get a chance, a quick peak for code-sanity?

It introduces a new step in the build process, and includes a view which displays build details.

Its for this story [Figure out version/build numbering …](https://www.pivotaltracker.com/story/show/109541948)

The PT story has a screenshot and a little background info.


dougmartin:
:+1: Looks good to me.  We may want to think about replacing build.sh with the gh-pages package Sam used on Cloud File Manager gulp deploy task.  It is a lot more succinct.

https://github.com/concord-consortium/cloud-file-manager/blob/master/gulp/tasks/deploy.js

knowuh:
deploy.js on cloud-file-manager certainly is concise, but it might not address our needs. We use deploy.sh to sometimes sometimes deploy to AWS cloudfront, also stuff ENV vars into the build depending on answers from prompts.  Part of this is for rollbar config.  There might be gulp tasks for those items too though. Next time we have deploy issues, we should dig into it.

sfentress:
Looks good to me. :+1:

PR #113, opened 2016-01-21
qualitative variables plus standalone features

bfinzer:
- qualitative for variables to be plotted
- enable mini-graph button so that it makes CODAP graph


knowuh:
I gave this a read. It looks good to me.  :+1:

PR #114, opened 2016-01-28
Use protocol-relative paths for CFM urls; avoid mixed-content errors.

knowuh:
Frieda encountered this tonight while trying to use Sage from   Lara(https) → Codap(https) → Sage(https).


knowuh:
Also: Mocha tests are failing presumably because we are trying out new simulation engines?

dougmartin:
Looks good :+1:

PR #127, opened 2016-02-10
Multi run steps

sfentress:


knowuh:
This looks like a nice simplifcation.  Removing math.js makes sense for now.  Adding it later (if we ever get to custom formulas) should be easy if we need it.  :+1:

sfentress:
To be clear, it's not gone, and it's still used if we try and solve an equation that has a string representation but no JS function associated with it. Which currently would only happen when we implement custom formulas, as you say.

At some point it might be nice to see if mathjs could have some way to cache the parsed equation, so it doesn't have to reparse each time. This seemed to be the real bottleneck.

dougmartin:
@sfentress I remember seeing this when looking at math.js for Teaching Teamwork:

http://mathjs.org/docs/expressions/parsing.html#parse

> An expression needs to be parsed and compiled only once, after which the expression can be evaluated repeatedly. On evaluation, an optional scope can be provided, which is used to resolve symbols and to write assigned variables or functions. Parameter scope is a regular Object.

I think you could write a wrapper to cache/memoize parse() and then eval it with the scope.  The docs show this:

    var node = math.parse(expr);      // parse expression into a node tree
    var code = node.compile();        // compile the node tree
    var result = code.eval([scope]);  // evaluate the code with an optional scope

That being said I'm not sure it is the parse/compile step that is taking all the time - it may be the code.eval step in which case it's moot.

sfentress:
Good find, thanks. I figured something like that must be available, but hadn't looked.

If we run into this slowdown with user-defined formulas, then we should definitely add this. As you say, though, the parsing may not be the slow part. (I didn't fully profile what was going on inside the `eval` function, I just saw that parsing called dozens of nested functions, but the result of that may have been quite fast.)

PR #129, opened 2016-02-12
Dependent sliders disabled

sfentress:
This PR sets sliders on dependent nodes to be disabled while model is running.


knowuh:
Code looks good to me, and I verified it works locally.  :+1:  @ddamelin

PR #143, opened 2016-03-09
Make semiquant proportional

sfentress:
@knowuh or @dougmartin if one of you would like to look at this that would be great.

This is already deployed to staging, so no rush.


knowuh:
Looks good to me @sfentress :+1:

PR #144, opened 2016-03-14
Switch data versions to use SemVer

sfentress:
@knowuh This updates all our old migration numbers to SemVer.

I can't see a problem with updating our old migration numbers. We convert the versions in the saved doc to their equivalent semvers before deciding to update, so we shouldn't run any migrations twice. However, I may be missing something.

### DIFF
--- a/src/code/data/migrations/migration-mixin.coffee
+++ b/src/code/data/migrations/migration-mixin.coffee
@@ -1,10 +1,15 @@
 # Implement version: xx and doUpdate: (data) ->  in your migrations.
 # and mixin this module
 
+semver = require "semver"
+
 module.exports =
-  # TODO: possibly use semver https://github.com/npm/node-semver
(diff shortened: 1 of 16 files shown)


src/code/data/migrations/migration-mixin.coffee
```diff
@@ -17,3 +22,12 @@ module.exports =
     else
       log.info "  skipped : #{@name()}"
     data
+
+  # Change x.y to "x.y.0". The only annoyance is we have to special-case 1.95,
+  # as this was supposed to be < 1.10.0
+  _semverize: (v) ->
```
knowuh: This looks like it handles converting the legacy floating point version numbers. I was worried about it, but it looks good.  Have you smoke tested against some saved models? Thanks for taking care of this.

knowuh:
Looks great, thanks.

PR #145, opened 2016-03-15
Add minigraphs to nodes

sfentress:
With this change, the `GraphStore` listens to the simulation events, and pushes frame data into
each `Node` object when we have new simulation data.

With this data, we can swap out the internal node image for a new `node-svg-graph-view` when the minigraph checkbox is checked.

The minigraph state is loaded and serialized, but the frame data is not.

Note that we see some similar speed issues as CODAP when we manipulate the slider while the
model is running. Now that this effect is local, we can more easily profile what's going on.

@knowuh or @dougmartin can you take a look sometime when you have a moment?


knowuh:
Looks good @sfentress :+1:

PR #161, opened 2016-04-06
Link styling - dashes for undefined, colors for vector, thickness for scale

dstrawberrygirl:


src/code/models/relation-factory.coffee
```diff
@@ -57,6 +63,7 @@ module.exports = class RelationFactory
     text: tr "~NODE-RELATION-EDIT.LESS_AND_LESS"
     postfixIco: "less-and-less"
     formulaFrag: "21.7 * log(in+1)"
+    magnitude: 1
```
sfentress: Should "less and less" be smaller than "a little?" Or in any case smaller than "about the same?"

PR #166, opened 2016-04-13
Hide dependent sliders

sfentress:
This

1. recognizes if nodes are in a cycle
2. shows enabled sliders before running a simulation for independent nodes, nodes in a cycle, and collectors
3. grays out sliders for nodes in a cycle and collectors during simulation
4. considers any node with only undefined inbound links as being a independent node

and tests are added.

https://www.pivotaltracker.com/story/show/116538381
https://www.pivotaltracker.com/story/show/117075519

### DIFF
--- a/src/code/models/node.coffee
+++ b/src/code/models/node.coffee
@@ -106,11 +106,14 @@ module.exports = class Node extends GraphPrimitive
   inNodes: ->
     _.map @inLinks(), (link) -> link.sourceNode
 
-  isDependent: ->
-    for link in @inLinks()
-      if link.relation and link.relation.isDefined
-        return false
-    return true
+  isDependent: (onlyConsiderDefinedRelations) ->
+    if onlyConsiderDefinedRelations
+      for link in @inLinks()
+        if link.relation and link.relation.isDefined
+          return true
+      return false
+    else
+      @inLinks()?.length > 0
 
   checkIsInCycle: ->
     visitedNodes = []
@@ -118,7 +121,7 @@ module.exports = class Node extends GraphPrimitive
 
     visit = (node) ->
       visitedNodes.push node
-      for link in node.outLinks()
+      for link in node.outLinks() when link.relation?.isDefined
         downstreamNode = link.targetNode
         return true if downstreamNode is original
         unless _.contains visitedNodes, downstreamNode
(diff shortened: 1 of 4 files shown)


src/code/models/node.coffee
```diff
@@ -160,8 +180,11 @@ module.exports = class Node extends GraphPrimitive
       frames: @frames
     key: @key
 
+  canEditInitialValue: ->
+    @isDependent() or @isAccumulator or @isInCycle
+
   canEditValueWhileRunning: ->
-    @inLinks().length is 0 or @isAccumulator
+    @isDependent()
```
knowuh: Is this inverted?  (just reading, its seems counter intuitive)

knowuh:
Super nice looking tests.  👍   I had one question about the name / function / my expectations about  `isDependent` but  looks great otherwise.

src/code/models/node.coffee
sfentress: Good eyes @knowuh. I had missed your comment, but had inverted that check anyway during a rebase.

sfentress:
@ddamelin For all loops.

To be honest, these kinds of tricky questions (are we really certain that initial values don't matter within a loop when there is an input, for *all* possible loop topologies?) are what makes me think that none of this work is going to survive. I'm fairly convinced that users are going to be so confused by the conditions that make sliders show or not show or gray that we're going to want to scrap the whole thing ;). But I'm happy to keep running with this algorithm and view choices for as long as possible.

ddamelin:
@sfentress I'm concerned about that too, which is why I'm trying to make it what I was thinking would be somewhat simple. If a loop is part of some larger network (which will almost always be the case) then we only show independent variable sliders. In the case where there is no way to influence the loop (it is self-contained and isolated) then we have to give some ability to change the sliders, because in some cases it will matter what the starting points are. 

I'm also open to hearing alternatives if something is occurring to you that would be better.

sfentress:
@ddamelin I can see that, but I get worried about the cases where we ought to show sliders, but have mistakenly calculated the loop as being part of a larger network, and therefore hidden them, preventing the user from affecting the output.

Take this example: https://dl.dropbox.com/s/y0lngo9yrj3zojn/Screenshot%202016-04-14%2009.48.01.png

We have two very similar pairs of loops, but they need to be treated differently. The A loop directly feeds into the B loop, so A should have sliders but B should not. But the C loop doesn't directly feed into D, as they share a node, so *all* of C and D should show sliders.

I think I know how to analyze the network to separate out those two cases, but I don't know (yet) what an algorithm would look like that would correctly cover all cases.

I'd be happy to spend more time on that, possibly with you, coming up with other topologies and working out a general algorithm, if you thought it important.

ddamelin:
Let's pause on this for now, so you can move to the CFM stories. I do see your point that it will likely be difficult to capture all use cases for potential intertwined loops. However, we may be able to come up with an algorithm that will deal with 95% of user models without too much complexity. Perhaps we can pull all the models from the recent classroom trial to see how the slider hiding algorithm would work for them.

PR #170, opened 2016-04-15
Fix call to clear selection when a link is removed

dstrawberrygirl:


sfentress:
:+1: Thank you for this one!

PR #190, opened 2016-05-24
Clamp less-and-less formula to prevent imaginary numbers

dstrawberrygirl:


sfentress:
This looks good to me, and the migration looks fine. 👍

PR #210, opened 2016-10-19
Upgrade CODAP-connect to API v2.0

sfentress:
@jsandoe I don't know if you have time to glance at this, but if you did that would be useful, thanks.

### DIFF
--- a/src/code/models/codap-connect.coffee
+++ b/src/code/models/codap-connect.coffee
@@ -38,18 +38,6 @@ module.exports = class CodapConnect
     @codapPhone = new IframePhoneRpcEndpoint( @codapRequestHandler,
       'data-interactive', window.parent )
 
-    # set frame size
-    @codapPhone.call
-      action: 'update',
-      resource: 'interactiveFrame',
-      values: {
-        title: @name,
-        dimensions: {
-          width: 800,
-          height: 600
-        }
-      }
-
     # load any previous data
     @codapPhone.call
       action: 'get',
--- a/src/code/models/codap-connect.coffee
+++ b/src/code/models/codap-connect.coffee
@@ -57,48 +57,56 @@ module.exports = class CodapConnect
       }
     ]
 
-    # create collections, then kick-off init
+    # check if we already have a datacontext (if we're opening a saved model).
+    # if we don't create one with our collections. Then kick off init
     @codapPhone.call
-      action: 'create',
-      resource: 'dataContext',
-      values: {
-        name: 'Sage Simulation',
-        title: 'Sage Simulation'
-        collections: [ {
-          name: 'Simulation',
-          title: 'Sage Simulation',
-          labels: {
-            singleCase: 'run',
-            pluralCase: 'runs'
-          },
-          attrs: [
+      action: 'get',
+      resource: 'dataContext[Sage Simulation]'
+    , (ret) =>
+      if ret?.success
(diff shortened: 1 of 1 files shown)


jsandoe:
Looks good to me.

 One thing I noticed, not directly related to the conversion, is that the 'run' attribute resolves to a formula (caseIndex). While this obviously works, I think it would be better if it were an explicitly set value, managed as a part of Sage Modeler's state. The reason for this is that it will be dynamically recalculated on the CODAP side if the user deletes the cases for a run.  If they are comparing runs in a graph, it could be confusing. I would think of a run number as permanent, once assigned. As I said, unrelated, but if there is an opportunity to fix, it might be worth-while.

ddamelin:
One of the reasons we may want to keep it caseIndex is because we would like to use the feature in the graphs that shows when app=is allowing the user to quickly change what is visible in the graph.

sfentress:
@ddamelin Can you explain further? I was just changing our API to match @jsandoe's comment.

jsandoe:
Ah, I see. The code that displays the inquiry space toggles just increments
through the interval defined by the number of parent cases, so this is what
@ddamelin is concerned about. There is no direct correlation between the
attribute defined by the caseIndex and the numeric toggle. It relies on the
attribute and the toggles staying in sync by virtue of the behavior of
caseIndex. We recently have had discussions about removing the "app=is"
query parameter and making the toggle behavior controllable as a property
of the data set. I think, when we do that we could explicitly bind the
symbol displayed for the toggle to an attribute in the table. Then, we
could consider removing this use of caseIndex.

Jonathan

 [image: Inline image 1]

sfentress:
Ok, since I'm out of the loop here and don't quite understand the implications, I'll just ask: what should I do here? Leave it as-is for now, or maintain our own record of Run Number and set it manually?

If this is something that's going to change at some point in CODAP, I have no problem making a change now, and a different change later.

ddamelin:
@sfentress I would leave it for now. If we do end up binding the display toggles in the graph to an attribute value, then we could change our code to generating a run ID, but we can't do that now without the other change being made.

sfentress:
@jsandoe I've updated this PR with the check to see if we already have a Data Context of the same name before creating one.

sfentress:
@jsandoe Just in case you don't get PT notifications from the BuildingModels project, this change relates to this bug: https://www.pivotaltracker.com/story/show/132775871

sfentress:
@jsandoe I just added code in `get/interactiveFrame` to listen to see if externalUndo is available (and, further in the future, standalone mode). (82ef60a)

Does this match what we discussed?

If so, I think this PR may be ready, pending the CODAP changes.

jsandoe:
@sfentriss, it does indeed. Change is up on
http://codap.concord.org/~jsandoe/build. I had not made the connection with
standalone mode, but that makes sense. So, CODAP would put that in the
interactiveFrame whenever it was invoked in standalone mode. Are there any
other considerations? Just a passthrough?

sfentress:
@knowuh I think that this can be merged, even without waiting for the CODAP changes to be in production. We just don't want to deploy it to production until CODAP has done so first. But merging it in now would probably make it easier to work on the changes to sending data to CODAP, so that those changes can use the new API, rather than writing it twice.

knowuh:
Thanks @sfentress  I will merge.

PR #211, opened 2016-10-25
New record buttons

knowuh:
@sfentress  If you could have a look.  

Most of this work is just cosmetic.  The largest changes are from the addition of a two Font Awesome glyphs into our font set.

The rest is changes to the simulation-run-panel.  It adds the three buttons for recording 1 datum, recording a stream, and recording for one period (of N intervals).

It's not wired yet to save data to CODAP.


sfentress:
This all looks good to me. Two minor things: The Record Data Stream button changes size when it changes state, maybe it should have a fixed size? And Travis is pointing out that there is a failing test.

Other than that, :+1:

knowuh:
@sfentress thanks for checking.  I fixed the failing test and the `record stream` button size doesn't change anymore when the text changes.

sfentress:
Looks good.

PR #212, opened 2016-10-26
Vertical slider

sfentress:
This adds a thermometer-style vertical slider to the right, and moves the two existing action circle.

The slider view now has many more options for horizontal, vertical, fillable and not. This is probably useless, as we are no longer using it anywhere else, now that the speed slider has gone, but we might as well keep it until maintaining these options becomes too painful.

The slider is shorter than what @ddamelin's mockups showed, as the apparent intent in the mockups was to make the thermometer level to be exactly aligned with the minigraph.

[#132117711] https://www.pivotaltracker.com/story/show/132117711
[#132832351] https://www.pivotaltracker.com/story/show/132832351
[#132117775] https://www.pivotaltracker.com/story/show/132117775

### DIFF
--- a/src/code/views/value-slider-view.coffee
+++ b/src/code/views/value-slider-view.coffee
@@ -287,4 +287,4 @@ Demo = React.createClass
         onRangeChange: @onRangeChange
     )
 
-# window.testComponent = (domID) -> ReactDOM.render React.createElement(Demo,{}), domID
+window.testComponent = (domID) -> ReactDOM.render React.createElement(Demo,{}), domID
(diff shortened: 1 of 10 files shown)


src/code/views/value-slider-view.coffee
```diff
@@ -32,8 +38,10 @@ ValueSlider = React.createClass
     "editing-min": false
     "editing-max": false
 
-  updateValue: (xValue,dragging) ->
-    value = @valueFromSliderUI(xValue)
+  updateValue: (locValue,dragging) ->
+    console.log("locValue = "+locValue)
```
knowuh: Should we remove these console messages?

PR #213, opened 2016-10-26
Recording button css fixes

knowuh:
@sfentress take a look?

These are mostly small CSS tweaks based on Demo meeting feedback.


sfentress:
It looks fine, although I wonder if the fact that the Stream button still switches to saying "Stop Recording" will trigger Cynthia's concern that it looks like you should be able to do something with that button. Maybe it should just be disabled, without changing the wording. (e.g. in simulation-run-panel-view:102 >> `if @state.isRecording and !@state.isRecordingOne`)

Either way, :+1:

PR #214, opened 2016-11-02
Live simulation mode

knowuh:
This is the start of work that decouples data recording (sending to CODAP) from simulation mode.

Note: This branch is failing some of the simulation tests at the moment. I wanted to get this PR in before I started work on the work related to recording data (sending to CODAP). I though that the PR would get too big otherwise.

Related Stories:

https://www.pivotaltracker.com/story/show/132118699
https://www.pivotaltracker.com/story/show/132117521
https://www.pivotaltracker.com/story/show/133207645
https://www.pivotaltracker.com/story/show/132117775

This PR also includes minor restyling of the 'because' link properties rational text-area:
https://www.pivotaltracker.com/story/show/133221703

### DIFF
--- a/src/code/stores/simulation-store.coffee
+++ b/src/code/stores/simulation-store.coffee
@@ -64,20 +64,21 @@ SimulationStore   = Reflux.createStore
 
   onExpandSimulationPanel: ->
     @settings.simulationPanelExpanded = true
+    @settings.modelReadyToRun = true
+    @settings.modelIsRunning = true
     @notifyChange()
 
   onCollapseSimulationPanel: ->
     @settings.simulationPanelExpanded = false
-    SimulationActions.simulationEnded()
+    @settings.modelReadyToRun = false
+    @settings.modelIsRunning = false
     @notifyChange()
 
   onGraphChanged: (data)->
     @nodes = data.nodes
     @settings.modelIsRunnable = @_checkModelIsRunnable()
     @settings.graphHasCollector = @_checkForCollectors()
-    if @settings.modelIsRunning
-      @_runSimulation()
-    @notifyChange()
+    @_runSimulation()
 
   onSetDuration: (n) ->
     @settings.duration = Math.max 1, Math.min n, 5000
@@ -109,6 +110,7 @@ SimulationStore   = Reflux.createStore
     if @settings.modelIsRunnable and @settings.modelReadyToRun
       # graph-store listens and will reset the simulation when
       # it is run to clear pre-saved data after first load
+      @settings.modelIsRunning = true
       duration = 1
       if @settings.graphHasCollector
         duration = @settings.duration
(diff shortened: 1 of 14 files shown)


knowuh:
@sfentress could you look at this if you get a moment.

src/code/views/link-relation-view.coffee
```diff
@@ -167,21 +167,30 @@ module.exports = LinkRelationView = React.createClass
...
+            xAxis: source
+            yAxis: target
+            link: @props.link
+            graphStore: @props.graphStore
+          )
         )
-      (div {className: 'fixed-bottom'},
+      )
+
+      (div {className: 'bottom'},
```
sfentress: The change from `fixed-bottom` to `bottom` makes it so that the graph and text box move when the second pull-down is added. Previously, this was adjusted so that those elements remained fixed in place at the bottom of the panel.

(I am to blame for overloading various variations on the classes named "bottom." It's the effect of making one incremental change after another and not cleaning them up...)

src/code/stores/simulation-store.coffee
```diff
@@ -66,12 +68,15 @@ SimulationStore   = Reflux.createStore
...
     @settings.simulationPanelExpanded = false
+    SimulationActions.simulationEnded()
     @notifyChange()
 
   onGraphChanged: (data)->
     @nodes = data.nodes
     @settings.modelIsRunnable = @_checkModelIsRunnable()
     @settings.graphHasCollector = @_checkForCollectors()
+    if @settings.modelIsRunning
+      @_runSimulation()
```
sfentress: While it seems to work really well for the most part for `@_runSimulation` to be called on GraphChanged, there seem like there are a couple of edge-cases where this shouldn't happen, namely: adding and removing links without defined relationships (since this doesn't actually change the underlying graph). I don't know whether it would be possible for making a new link without a relationship to not call `onGraphChanged`, or if some other filter is needed.

The other place where this is trouble is undo/redo. If you move a slider, the simulation is run, and then if you undo that movement, this still triggers new data to be added, so the graph keeps moving forward. It seems like this would need to be special-cased some way, to correctly allow for undoing data additions, and maybe that's too much for this PR's scope.

src/code/views/node-view.coffee
```diff
@@ -165,15 +165,18 @@ module.exports = NodeView = React.createClass
 
   renderSliderView: ->
     enabled = not @props.running or @props.data.canEditValueWhileRunning()
+    showHandle = @props.data.canEditInitialValue()
+    value = @props.data.currentValue or @props.data.initialValue
```
sfentress: The annoyingly-common JS bug: zero will count as false here, so if you have a dependent node and you bring it all the way down to zero, you'll see the fill slider value pop back up to initialValue.

Suggest change to `value = @props.data.currentValue ? @props.data.initialValue`

src/code/views/link-relation-view.coffee
knowuh: This change intentionally changes the fixed-bottom nature of the text box.  Dan thought it was better to have predictable spacing between the because box and graph, as per this story:  https://www.pivotaltracker.com/story/show/133221703  -- in cases where variable names are long the graph and because box would overlap.  Dan thought it would be better to scroll in such instances.

src/code/views/node-view.coffee
knowuh: Thank you!  Good catch!

PR #218, opened 2016-11-10
Record simulation changes

knowuh:
An in-progress branch.

'Simulation' is been decoupled from sending to CODAP in this branch.  It also includes an improvement to how we are listening to simulation changes in the simulation tests.

Here are some related PT stories:

User can collect a single data point and send it to CODAP when there are no collectors.
https://www.pivotaltracker.com/story/show/133207443

User can turn on "recording" to send a stream of data when no collectors are present.
https://www.pivotaltracker.com/story/show/133207187

### DIFF
--- a/src/code/stores/simulation-store.coffee
+++ b/src/code/stores/simulation-store.coffee
@@ -13,9 +13,9 @@ SimulationActions = Reflux.createActions(
     "resetSimulation"
     "setDuration"
     "setStepUnits"
-    "setSpeed"
     "simulationStarted"
     "simulationFramesCreated"
+    "recordingFramesCreated"
     "simulationEnded"
     "capNodeValues"
     "recordStream"
@@ -78,7 +78,9 @@ SimulationStore   = Reflux.createStore
     @nodes = data.nodes
     @settings.modelIsRunnable = @_checkModelIsRunnable()
     @settings.graphHasCollector = @_checkForCollectors()
-    @_runSimulation()
+    # TODO:  This can't go here: We want this for 'real time' simulations
+    # But we need something with better granularity.
+    # @_runSimulation()
 
   onSetDuration: (n) ->
     @settings.duration = Math.max 1, Math.min n, 5000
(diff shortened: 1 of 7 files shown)


knowuh:
@sfentress -- this is in progress, but please have a look before it gets too big.

I will start work on a second PR where we remove 'speed'.

Thanks!

sfentress:
Should I be able to run this? I'm getting errors about `stripTiming` not being defined.

knowuh:
@sfentress you should be able to run it.  If not I must have local changes not in the branch.

knowuh:
@sfentress  -- sorry you are right -- I have local changes here. Let me sort it out.

knowuh:
sorry about that @sfentress  -- when you get a chance, it should run now :)

src/code/stores/simulation-store.coffee
```diff
@@ -101,6 +101,12 @@ SimulationStore   = Reflux.createStore
   onRunSimulation: ->
     @_runSimulation()
 
+  stepUnits: () ->
+    if @isRecordingPeriod
```
sfentress: Should this be `@settings.isRecordingPeriod`? Non-step time units don't seem to get sent, otherwise.

src/code/stores/simulation-store.coffee
```diff
@@ -115,10 +121,19 @@ SimulationStore   = Reflux.createStore
...
+        # Simulation events get triggered as Actions here, and are
+        # available to anyone who listens to this store
         onFrames: (frames) =>
           SimulationActions.simulationFramesCreated(frames)
           if @settings.isRecording
-            SimulationActions.recordingFramesCreated(frames)
+            if @settings.isRecordingPeriod
+              SimulationActions.recordingFramesCreated(frames)
+            else
+              # Strip out the time information if we are simulating without time...
```
knowuh: Yes, it doesn't seem like that has made it into the current API yet. (I saw that its committed -- but not deployed)

src/code/stores/simulation-store.coffee
knowuh: Thanks. Let me double check that.

knowuh: Thanks @sfentress -- I totally missed that.  Running c07352c locally I now get new columns in CODAP when I change units.   

(we don't really want to add new columns, but that is a different story)

PR #220, opened 2016-11-16
Replace jQuery slider with pure React

sfentress:
@knowuh Could you take a look at this?

I was hoping that we'd be able to get rid of the jQuery dependency entirely, but it looks like we're still using it in a number of places. Maybe one day we'll get an intern to go through all the references armed with http://youmightnotneedjquery.com/

### DIFF
--- a/src/code/views/value-slider-view.coffee
+++ b/src/code/views/value-slider-view.coffee
@@ -78,11 +78,11 @@ ValueSlider = React.createClass
     window.removeEventListener 'resize', @handleUpdate
 
   handleUpdate: ->
-    { orientation } = this.props
+    { orientation } = @props
     dimension = constants.orientation[orientation].dimension
     dimension = dimension.charAt(0).toUpperCase() + dimension.substr(1)
-    sliderPos = this.slider["offset#{dimension}"]
-    handlePos = this.handle?["offset#{dimension}"] or 0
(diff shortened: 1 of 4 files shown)


knowuh:
This looks much improved.  Thanks @sfentress !

sfentress:
I will add a fix to solve the issue where the handle goes outside the range if it's set to an out-of-bounds value.

src/code/views/value-slider-view.coffee
```diff
@@ -125,34 +111,95 @@ ValueSlider = React.createClass
...
+    e.preventDefault()
+
+  handleStart: (e) ->
+    @handleNoop(e)
+    document.addEventListener 'mousemove', @handleDrag
+    document.addEventListener 'mouseup',   @handleEnd
+
+  handleEnd: ->
+    document.removeEventListener 'mousemove', this.handleDrag
+    document.removeEventListener 'mouseup',   this.handleEnd
```
sfentress: No reason at all. I'll clean that up with my other changes.

sfentress:
@knowuh I've updated this to to fix the two bugs, slider going out of bounds and slider dragging a selected node.

https://www.pivotaltracker.com/story/show/134443309
https://www.pivotaltracker.com/story/show/134443393

knowuh:
This looks fine @sfentress -- thanks

PR #221, opened 2016-11-16
Add cache-busting query to app.js and app.css

sfentress:
@knowuh This is a straw-man idea for cache-busting, using query parameters on app.js and app.css that are unique per deployment (i.e. request `app.js?v={COMMIT_SHA}`).

There may well be other methods of dealing with this. Thoughts/conversation appreciated.

https://www.pivotaltracker.com/story/show/134090627

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -12,21 +12,23 @@ var argv = require('yargs').argv,
       tag: gitTags.split('-')[0],
       commiter: gitLog.split(" ")[2],
       commit: gitLog.split(" ")[1]
-    };
+    },
     buildInfoString = (buildInfo.date + " " +
       buildInfo.tag + " " +
       buildInfo.commit + " " +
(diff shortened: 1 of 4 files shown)


knowuh:
@sfentress seems good to me. I was originally worried because I thought that the presence of a query parameter would always cause a full download of assets.  But this SO Q&A convinced me that this is probably fine: http://stackoverflow.com/questions/9692665/cache-busting-via-params

It seesm like we could completely invalidate the cloud-front cache when deploying by changing some options in `s3_website.yml`, as specified [in the readme](https://github.com/laurilehmijoki/s3_website#invalidating-all-cloudfront-resources-wildcard-invalidation) -- It looks like specifying both `cloudfront_invalidate_root: true` and `cloudfront_wildcard_invalidation: true` would help us..

But clearly this approach should work great too.

PR #222, opened 2016-11-17
Rm simulation speed

knowuh:
This PR should be reviewed after Recording / Simulation PR is merged. 

99aaed3 removed speed from the simulation.
a57e5e4 add migrations to remove speed from data & modifies tests
99aaed3 simply silences coffelint warnings.


sfentress:
I think this all looks fine. :+1:

PR #224, opened 2016-11-18
Some simulation panel fixes

knowuh:
These commits address state bugs change in the simulation panel.

* should update the recording buttons in the simulation panel when the diagram starts using collectors.
* should stop recording when the diagram gets a collector.
* should stop recording when the simulation panel is closed.

If you have a second to review @sfentress 

Thanks!


sfentress:
This all looks good.

PR #225, opened 2016-11-18
Gray mini graphs and slider bars when not recording.

knowuh:
This PR lets the data colors for the mini-graph and the sliders be adjusted dynamically.

When the simulation is being recorded, the data is displayed in CODAP orange.

Otherwise is gray.

@sfentress please review when you get a chance.

### DIFF
--- a/src/code/utils/colors.coffee
+++ b/src/code/utils/colors.coffee
@@ -1,12 +1,25 @@
 tr = require './translate'
-module.exports =
-  [
+
+colors =
+  yellow:
     name: tr("~COLOR.YELLOW")
     value: "#f7be33"
-  ,
+  darkBlue:
     name: tr("~COLOR.DARK_BLUE")
     value: "#105262"
-  ,
+  medBlue:
     name: tr("~COLOR.MED_BLUE")
     value: "#72c0cc"
-  ]
+  lightGray:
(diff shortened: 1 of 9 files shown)


knowuh:
@sfentress -- not sure why this is failing CI tests, local tests seem to be passing.

I think this is ready for review now.  Thanks!

src/code/utils/colors.coffee
```diff
@@ -1,12 +1,25 @@
...
+  darkGray:
+    name: tr("~COLOR.DARK_GRAY")
+    value: "#444"
+  data:
+    name: tr("~COLOR.DATA")
+    value: "#e99373"
+
+module.exports =
+  colors: colors
+  choices: [colors.yellow, colors.darkBlue, colors.medBlue]
```
sfentress: I don't quite understand the need for `Colors.choices`. Is it just so that you can use `for color in Color.choices` in color-picker-view? Because without it, you could still use `for color of Colors`.

As it is, the use of `Color.colors.[...]` everywhere just reads a little awkward.

sfentress: Oh, wait, I see now that `choices` is a subset of all the colors. Ok, I get it now. Ignore this.

PR #227, opened 2016-11-21
Graph codap improvements

sfentress:
Small graph fixes:

1. Show graph button only when in CODAP
2. Don't let button disappear when user hovers over link button
3. Set correct x-axis units if we have selected some units for a collector experiment, and then open a graph

https://www.pivotaltracker.com/story/show/132771013  
https://www.pivotaltracker.com/story/show/134408727  
https://www.pivotaltracker.com/story/show/133192489


knowuh:
Looks good!

PR #229, opened 2016-11-22
Make "a little" relation balanced around midpoint

sfentress:
@knowuh Another small one.

In "increase" direction, this goes from 1/4 max to 3/4 max, and vice-
versa in other direction.

Since we're dealing with formula fragments, we can't just add an
intercept, since it changes based on vector, so this is a horizontal
translation.

[#133219491]


knowuh:
Looks good.  And the simulation follows my intuition. 👍

PR #230, opened 2016-11-23
Simulation should run when impactful diagram changes made.

knowuh:
Simulation state should update whenever a change is made that would impact the values of the variables.
    * relationship change
    * deleted link
    * range change on slider

[#134471467]
https://www.pivotaltracker.com/n/projects/1263626/stories/134471467

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -119,6 +119,7 @@ GraphStore  = Reflux.createStore
     @undoRedoManager.createAndExecuteCommand 'addLink',
       execute: => @_addLink link
       undo: => @_removeLink link
+    SimulationStore.actions.runSimulation()
 
   _addLink: (link) ->
     unless link.sourceNode is link.targetNode or @hasLink link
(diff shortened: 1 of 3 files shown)


src/code/stores/graph-store.coffee
```diff
@@ -119,6 +119,7 @@ GraphStore  = Reflux.createStore
     @undoRedoManager.createAndExecuteCommand 'addLink',
       execute: => @_addLink link
       undo: => @_removeLink link
+    SimulationStore.actions.runSimulation()
```
sfentress: If we're trying to run the simulation only when the simulation graph changes topology, there are a few events here that trigger `runSimulation` even when no change has been made. At minimum, this is generating redundant runs when a link is added with no relation; when a link without a relation is deleted; when a user makes a selection on the first pull-down menu when defining a relation; and when a user edits a link's name.

`addLink` won't ever contain a defined relation (if it's done via the UI), and so it seems like this one here could be removed.

Perhaps for the other two, `removeLink` on 138 and `changeLink` on 291, perhaps they could be prepended with `if link.relation.isDefined`?

This wouldn't resolve the link name-change run, so maybe filtering of properties would be needed for that (like you do in `changeNode` with `data.isAccumulator?`).

PR #232, opened 2016-11-23
Dynamically pluralize unit names if simulation length > 1

knowuh:
@sfentress maybe when we get back from break, you can take a look at this one.

[#133194843]
https://www.pivotaltracker.com/story/show/133194843

### DIFF
--- a/src/code/stores/simulation-store.coffee
+++ b/src/code/stores/simulation-store.coffee
@@ -83,12 +83,12 @@ SimulationStore   = Reflux.createStore
     @notifyChange()
 
   _updateUnitNames: ->
-    if @settings.duration > 1
-      @settings.timeUnitOptions = ({name: TimeUnits.toString(unit, true), unit: unit} for unit in TimeUnits.units)
-      @settings.stepUnitsName = TimeUnits.toString(@settings.stepUnits, true)
-    else
+    if @settings.duration is 1
       @settings.timeUnitOptions = ({name: TimeUnits.toString(unit, false), unit: unit} for unit in TimeUnits.units)
       @settings.stepUnitsName = TimeUnits.toString(@settings.stepUnits, false)
+    else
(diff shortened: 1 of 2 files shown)


kswenson:
@knowuh I'm not familiar with the full context here, but having written pluralization code in the past, an oddity of pluralization in English is that < 1 also uses the plural form, e.g. "0.5 seconds" or even "0 seconds". In the situations I've dealt with the code usually ends up looking something like `value === 1 ? singularForm : pluralForm`.

knowuh:
@kswenson -- even though we won't be using fractional units, I like your approach.  I will change it.

src/code/stores/simulation-store.coffee
```diff
@@ -81,13 +82,22 @@ SimulationStore   = Reflux.createStore
...
     @notifyChange()
 
+  _updateUnitNames: ->
+    if @settings.duration is 1
+      @settings.timeUnitOptions = ({name: TimeUnits.toString(unit, false), unit: unit} for unit in TimeUnits.units)
+      @settings.stepUnitsName = TimeUnits.toString(@settings.stepUnits, false)
+    else
+      @settings.timeUnitOptions = ({name: TimeUnits.toString(unit, true), unit: unit} for unit in TimeUnits.units)
+      @settings.stepUnitsName = TimeUnits.toString(@settings.stepUnits, true)
+
```
sfentress: This seems like it could be simplified to something like
```
pluralize = @settings.duration isnt 1
@settings.timeUnitOptions = ({name: TimeUnits.toString(unit, pluralize), unit: unit} for unit in TimeUnits.units)
@settings.stepUnitsName = TimeUnits.toString(@settings.stepUnits, pluralize)
```
just because I worry that the repetition might lead to errors changing something in one place and forgetting it in the other.

PR #234, opened 2016-11-29
Explicit experiment button

knowuh:
Adds a button and counter at the top of the simulation panel that lets a student specify the start of a new experiment.

[PT story for expiriments](https://www.pivotaltracker.com/story/show/134408545)

[PT story for not creating cases until record pushed](https://www.pivotaltracker.com/story/show/135145637)

@sfentress I think this is ready for a review if you get a chance.

### DIFF
--- a/src/code/models/codap-connect.coffee
+++ b/src/code/models/codap-connect.coffee
@@ -30,9 +30,8 @@ module.exports = class CodapConnect
     @sendThrottleMs = 300
 
     # Create a new case in these instances:
-    SimulationStore.actions.resetSimulation.listen         @_openNewCase.bind(@)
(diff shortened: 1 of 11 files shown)


sfentress:
Hi @knowuh,

just looking at the behavior when testing it, I notice that I can often create multiple experiments with the same number. If I create a model, press "Run Stream", and start moving the handle, I'll frequently get multiple "Experiment 1" parent rows.

Also, more minor, if I create a valid graph, with a defined relationship, and then open the simulate panel for the first time, the buttons start out gray until I change something else.

knowuh:
Thanks @sfentress  -- 33676f8 is the same fix we looked at together, plus a fix for the incorrectly disabled record buttons you mentioned.

sfentress:
Looks good. :+1:

PR #235, opened 2016-11-29
Run simulation synchronously

sfentress:
This eliminates the issue where the simulation runs multiple times with the same value, leading to straight lines in the graphs.

It also seems to have a positive effect on the feeling of "lag" when moving a slider, although I'm still trying to profile the speed accurately.


knowuh:
This looks good to me @sfentress

PR #236, opened 2016-12-01
Use diffing to decide when to draw jsPlumb, run simulation. Speed up model.

sfentress:
This adds "model descriptions," which are easily-comparable reductions of the model state to help decide when to redraw jsPlumb, or rerun the model.

By not redrawing jsPlumb everytime we run a simulation step, we improve our running speed significantly.

https://www.pivotaltracker.com/story/show/134467689

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -386,17 +386,22 @@ GraphStore  = Reflux.createStore
   toJsonString: (palette) ->
     JSON.stringify @serialize palette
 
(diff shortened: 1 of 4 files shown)


sfentress:
@knowuh Please take a look when you have time.

The last commit, "Only run simulation..." is not a critical piece of this, so if you think it needs more thought we can take it out. However, it seems to work, as far as I can tell.

knowuh:
🚀   FAST.

Thanks @sfentress  looks fine.

PR #237, opened 2016-12-01
Internal library images clean up

mpsota:
Those commits fix few minor things in Library Images modal.
- internal library search is removed
- few new images are added
- images that are already in the pallet are correctly greyed out.


knowuh:
@mpsota  This looks good to me. 👍

PR #238, opened 2016-12-05
Tweak z-indexes to render component correctly

mpsota:
Just 2 small changes:
- set line's z-index to 0,
- render palette on the top (z-axis = 3)


sfentress:
Looks good to me. :+1:

PR #239, opened 2016-12-21
Multiple nodes selection support

mpsota:
Those commits implement multiple nodes/links selection support:
- It is possible to select links and nodes with ctrl key
- It is possible to select links and nodes by drawing marquee
- More then one node can be moved at once
- More than one node/link can be removed at once.


mpsota:
I have few fixes to make. Please wait with the code review.

mpsota:
Selection works better now, review my changes please.

sfentress:
As far as I can tell, this looks all good. I think the code all looks clean.

I think the tests should be updated. It should be pretty straightforward to add tests to `selection-manager-test` that can cover the basic new functionality -- multiple node and link selection, and multiple deletion.

A more minor issue is that the marquee selection doesn't select a node unless you enclose the very top of the node -- i.e. if you draw a box from the bottom up, you can cover about 90% of the bottom and the node still won't be selected. Personally I wouldn't think this was blocking, though, if it's a tricky fix.

sfentress:
I think this looks good. Nice work. :+1:

PR #240, opened 2017-01-09
Add Yarn for improved dependency management.

knowuh:
Thanks for suggesting this @kswenson!   This was easier than I thought.  Am I missing anything?


README.md
```diff
@@ -24,11 +24,13 @@ we are using.
...
     * [Gulp](http://gulpjs.com/): `npm install -g gulp`
     * [live-server](https://www.npmjs.com/package/live-server) `npm install -g live-server`
+    * [yarn](https://yarnpkg.com/) `npm install -g yarn`
+`
 
 2. Check out the project and install the local dependencies:
     * `git checkout https://github.com/concord-consortium/building-models.git`
     * `cd building-models`
-    * `npm install`.
+    * `yarn`
```
dougmartin: I think this is `yarn install`

kswenson: `yarn` is a synonym for `yarn install`

kswenson:
Not missing anything that I'm aware of -- adding the `yarn.lock` file is all that's required. One thing to consider with respect to the ReadMe is that you may not want to imply that yarn is required as it potentially presents a (minor) barrier to those getting started. You could say `npm install` or `yarn`, for instance, since for most initial installs either should work just fine.

PR #241, opened 2017-01-10
Serialize Experiment number

knowuh:
This adds experimentNumber and experimentFrame to the serialized data. 

I had originally thought this data was ephemeral, but we really need to keep track of it to prevent overwriting previous cases, as described in the PT story: https://www.pivotaltracker.com/story/show/135554497

@sfentress could you take a look? 

[#135554497]

### DIFF
--- a/src/code/data/migrations/18_serialize_experiment_number.coffee
+++ b/src/code/data/migrations/18_serialize_experiment_number.coffee
@@ -1,0 +1,10 @@
+migration =
+  version: "1.17.0"
+  description: "Adds experiment number to serialization"
+  date: "2017-01-09"
+
+  doUpdate: (data) ->
+    data.settings.simulation?.experimentNumber ||= 0
+    data.settings.simulation?.experimentFrame ||= 0
+
+module.exports = _.mixin migration, require './migration-mixin'
--- a/src/code/data/migrations/migrations.coffee
+++ b/src/code/data/migrations/migrations.coffee
@@ -18,6 +18,7 @@ migrations = [
   require "./15_update_less_relationship_formula"
   require "./16_add_link_reasoning"
   require "./17_remove_simulation_speed"
+  require "./18_serialize_experiment_number"
 ]
 
 module.exports =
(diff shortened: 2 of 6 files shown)


sfentress:
The code all looks good. :+1:

One question: If you save a model in CODAP with data written to experiment 1, and then you re-open that model and add new data, it will add it to a new parent case, also named 1.

I don't know the CODAP API off the top of my head to know if it would be possible to query CODAP and see if a parent case with `Exp# = 1` already exists, but this might be nice if so.

Or -- and possibly less surprising for the user? -- you could auto-increment when re-opening a model with existing data. Although, thinking about it, this might not be possible to know unless we add yet-more data to the serialization.

In any case, I think is fine, if that behavior is seen as ok.

ddamelin:
If we are not going to write into the same existing experiment number, then we should automatically increment the experiment number upon loading the saved state.

sfentress:
... but presumably only if we've saved data into that experiment, which is the annoying part because that information is only in CODAP.

I think you could request a list of the parent cases, read the exp#'s, and select the next-highest?

knowuh:
I ended up changing a lot of  the `codap_connect.coffee`  

Much of the work is done in `_createMissingDataAttributes` which handles comparing attributes of the samples collection in CODAP to the current set of diagram nodes.

Another big change: instead of creating new child casses, we use the [`create item`](https://github.com/concord-consortium/codap/wiki/CODAP-Data-Interactive-API#items) CODAP message.  This allows us to handle the case where the parent case (with the current experiment number) has been deleted.

@sfentress if you could take a look.  If you don't have time, we could bring @jsandoe into the mix instead.

sfentress:
As far as I can tell, this looks good. I don't know whether it needs more testing at the edge-cases, but it seems like it makes sense to me.

It does seems as if there is a lot of repeated legwork in checking the attributes with every step, but as far as I can see it doesn't visibly impact the speed in any way, so I'm not sure if it's a problem.

:+1:

PR #242, opened 2017-01-12
Internal library images clean up

mpsota:
These commits improve the search results list.

- images already used are hidden in the search results,
- internal images are hidden when searching was started


knowuh:
Looks good 👍

PR #247, opened 2017-03-30
Double-click on link should now brings up relationship panel [#142637503]

dougmartin:
This removes the existing double click handler which was not being called anything this the link had a click handler.  Double clicking is detected with a simple test for two clicks within 250ms.

To enable the inspector panel to be commanded to open a panel its state was moved into a new store.


dougmartin:
@knowuh Good catch - Dan actually mentioned that in the planning meeting and I forgot about it.

dougmartin:
@knowuh Dan added double-clicking on a node also in a new story (https://www.pivotaltracker.com/story/show/142979121).  I've added a commit to handle it.

PR #251, opened 2017-04-03
Changed default recording unit to be days [#135136847]

dougmartin:
NOTE: The default quantity is still set at 50.  There was no mention in the story if that was to be changed so I left it alone.

To update the default quantity change this line: https://github.com/concord-consortium/building-models/blob/481f0579c63331c089dd08bda53dd56d19a1be6e/src/code/stores/simulation-store.coffee#L39

### DIFF
--- a/src/code/stores/simulation-store.coffee
+++ b/src/code/stores/simulation-store.coffee
@@ -24,6 +24,7 @@ SimulationActions = Reflux.createActions(
     "recordingDidStart"
     "recordingDidEnd"
     "createExperiment"
+    "toggledCollectorTo"
   ]
 )
 SimulationActions.runSimulation = Reflux.createAction({sync: true})
@@ -34,8 +35,9 @@ SimulationStore   = Reflux.createStore
     ImportActions, GraphActions]
 
   init: ->
-    defaultUnit = TimeUnits.defaultUnit
-    unitName    = TimeUnits.toString(defaultUnit,true)
+    @defaultUnit = TimeUnits.defaultUnit
+    @unitName    = TimeUnits.toString(@defaultUnit,true)
+    @defaultCollectorUnit = TimeUnits.defaultCollectorUnit
     defaultDuration = 50
     timeUnitOptions = ({name: TimeUnits.toString(unit, true), unit: unit} for unit in TimeUnits.units)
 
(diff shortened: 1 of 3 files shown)


ddamelin:
50 is fine for now

dougmartin:
@sfentress / @knowuh - this needs a new review - it was a bigger story than I understood it to be.

sfentress:
If the user has a graph with no collectors, and they change the time step to Hours, and then they toggle a node's collector checkbox on and off again, the simulation will reset to Steps. And similarly if they have a graph with a collector, a custom time step, and they toggle the collector off and on again. Is that behavior ok?

Another way to do it might be to to have `SimulationStore.collectorUnitName` alongside the existing `SimulationStore.unitName`, which would default at the start to `TimeUnits.defaultCollectorUnit`. Then when you switch between 0 and 1 collectors, you could switch between collectorUnitName and unitName, instead of switching between the defaults. Then you'd preserve the user's choice.

But if the behavior as shown is ok, then this looks ok. 

The only thing I wonder is whether it wouldn't be cleaner to use the existing `onGraphChanged` function in SimulationStore, instead of adding this single-purpose call to `onToggledCollectorTo`. The variable `@settings.graphHasCollector` could be changed to `@settings.numCollectors`, and the store is already using logic in the `_updateGraphHasCollector` method called by  `onGraphChanged` to see if the previous state == the current state, so the same thing could be done to see if you're toggling between 0 and 1. (...however, as I typed that out, I'm not sure that it's actually cleaner. It just removes the call to `toggledCollectorTo`, which seemed awkward. Your call.)

ddamelin:
I would prefer to toggle between the user's choice rather than reverting to the default timestep for collectors. (For all non-collector it will be "steps" regardless).

src/code/stores/simulation-store.coffee
```diff
@@ -34,8 +35,10 @@ SimulationStore   = Reflux.createStore
     ImportActions, GraphActions]
 
   init: ->
-    defaultUnit = TimeUnits.defaultUnit
-    unitName    = TimeUnits.toString(defaultUnit,true)
+    @defaultUnit = TimeUnits.defaultUnit
+    @unitName    = TimeUnits.toString(@defaultUnit,true)
+    @defaultCollectorUnit = TimeUnits.defaultCollectorUnit
+    @collectorUnitName    = TimeUnits.toString(@defaultCollectorUnit,true)
```
sfentress: The variable `@collectorUnitName` isn't used. Should it be used in `onSetStepUnits` and `onToggledCollectorTo` instead of `@defaultCollectorUnit`?

dougmartin:
I've removed the unused variable - that was just a copy-paste artifact

PR #253, opened 2017-04-03
Fixed missing initial link drawing when loading a saved file [#142963961]

dougmartin:
When using sage.html outside of CODAP the links where not being drawn at startup if a file was loaded via the command line like this: sage.html#file=localStorage:node%20test

This change forces the graph to draw when it is mounted.


dougmartin:
@sfentress There seems to be several CFM issue - the hash isn't updating sometimes and the document title isn't either.  I'll mention that at the demo today.

PR #254, opened 2017-04-03
Make deleting a node more intuitive when variable name editing is the focus. [#136380649]

dougmartin:
With this change the code now tracks if a node is added during the session (as opposed to loaded from a save) and then deletes the node if during the initial edit the first key pressed is either backspace or delete.

This also fixes the global key handler to also listen for delete (46) along with the existing backspace (8) listener.

### DIFF
--- a/src/code/views/node-view.coffee
+++ b/src/code/views/node-view.coffee
@@ -235,13 +235,12 @@ module.exports = NodeView = React.createClass
         max: @props.data.max
         data: @props.data.frames
         color: @props.dataColor
+        image: @props.data.image
       })
     else
       (SquareImage {
         image: @props.data.image,
-        ref: "thumbnail"
       })
-
 
   render: ->
     style =
--- a/src/code/views/node-view.coffee
+++ b/src/code/views/node-view.coffee
@@ -16,6 +16,10 @@ NodeTitle = React.createFactory React.createClass
     if @props.isEditing
       @inputElm().off()
 
+  componentWillUpdate: (nextProps) ->
+    # mark the title as updated even if no change was made when it leaves edit mode
+    @titleUpdated = true if @props.isEditing and not nextProps.isEditing
+
   componentDidUpdate: ->
     if @props.isEditing
       $elem = @inputElm()
@@ -33,7 +37,13 @@ NodeTitle = React.createFactory React.createClass
   inputValue: ->
     @inputElm().val()
 
+  detectDeleteWhenEmpty: (e) ->
+    # 8 is backspace, 46 is delete
+    if e.which in [8, 46] and not @titleUpdated
+      @props.graphStore.removeNode @props.nodeKey
+
   updateTitle: (e) ->
+    @titleUpdated = true
     newTitle = @cleanupTitle(@inputValue())
     @props.onChange(newTitle)
 
@@ -49,10 +59,12 @@ NodeTitle = React.createFactory React.createClass
 
   renderTitleInput: ->
     displayTitle = @displayTitleForInput(@props.title)
+    canDeleteWhenEmpty = @props.node.addedThisSession and not @titleUpdated
     (input {
       type: "text"
       ref: "input"
       className: "node-title"
+      onKeyUp: if canDeleteWhenEmpty then @detectDeleteWhenEmpty else null
       onChange: @updateTitle
       value: displayTitle
       maxLength: @maxTitleLength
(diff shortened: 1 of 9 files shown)


src/code/views/node-view.coffee
```diff
@@ -12,6 +12,9 @@ NodeTitle = React.createFactory React.createClass
   displayName: "NodeTitle"
   mixins: [require '../mixins/node-title']
 
+  componentWillMount: ->
+    @canDeleteWhenEmpty = @props.node.addedThisSession and not @props.node.titleUpdated
```
sfentress: Is it ever possible for `@props.node.titleUpdated` to be `true` here?

dougmartin: Yes, `titleUpdated` isn't set until a change occurs and it is noted on line 45 - before that is is `undefined`.  So `canDeleteWhenEmpty` is true when a node is added during the session and the user first edits the title (which happens automatically).  The blur event on this component sets `titleUpdated` to true so on subsequence edits of the title a backspace or delete doesn't delete it.

sfentress: > on subsequent edits of the title (which mounts this component)

node-view gets re-mounted when you edit the title? When testing, it doesn't seem to, so `@props.node.titleUpdated` looks like it can only ever be `undefined` at this point. But maybe I'm missing something.

dougmartin:
@sfentress Good catch.  I've moved the detection to the prop change handler and forced the updated flag to be true when the isEditing property transitions from true -> false.  The blur handler *isn't* being called when you click outside the input while editing so the forcing of the flag is needed when the user drops a new node and then doesn't change the title from the default and then clicks the background.

src/code/views/node-view.coffee
```diff
@@ -16,6 +16,11 @@ NodeTitle = React.createFactory React.createClass
     if @props.isEditing
       @inputElm().off()
 
+  componentWillUpdate: (nextProps) ->
+    # mark the title as updated even if no change was made when it leaves edit mode
+    @props.node.titleUpdated = true if @props.isEditing and not nextProps.isEditing
+    @canDeleteWhenEmpty = @props.node.addedThisSession and not @props.node.titleUpdated
```
sfentress: Another quick question just to understand this better: why have `@canDeleteWhenEmpty` be a class variable, but have `titleUpdated` be a prop? `titleUpdated` could just be a variable, right, since we only care about it during the lifecycle of this view?

sfentress: I feel like the `@canDeleteWhenEmpty` variable adds just a little bit of extra confusion to what's otherwise a simple check, and it takes an extra moment of thinking to see how it relates to a handler being added or not on every re-render of the view. It seems to me that it would just be a little simpler to simply have the check for both  `addedThisSession ` and `titleUpdated` in `detectDeleteWhenEmpty` (or maybe `detectDeleteIfNew`). 

For instance, you're already checking `titleUpdated` again in that method, which I *think* is now unnecessary, since the handler will be removed on the first rerender after the title is updated, but I feel like it shouldn't be necessary to work out whether the handler will always be removed at that point or not. 

So, simpler would just be to do both checks in the handler, and always have the hander attached.

But if you think this makes more sense to you, I don't feel too strongly about it.

dougmartin: You are right - when I added it I thought that the input was being mounted/unmounted whenever the node title was edited.  I see now that it always remains mounted so I will move that to a view variable.

dougmartin:
@sfentress I've moved the titleUpdated out of the props and moved the check for can delete down to the render handler.  It is much cleaner now, thanks!

src/code/views/node-view.coffee
```diff
@@ -49,10 +59,12 @@ NodeTitle = React.createFactory React.createClass
 
   renderTitleInput: ->
     displayTitle = @displayTitleForInput(@props.title)
+    canDeleteWhenEmpty = @props.node.addedThisSession and not @titleUpdated
     (input {
       type: "text"
       ref: "input"
       className: "node-title"
+      onKeyUp: if canDeleteWhenEmpty then @detectDeleteWhenEmpty else null
```
sfentress: I still don't get why we need `canDeleteWhenEmpty`, when you've got to test `@titleUpdated` twice anyway, instead of just checking `@props.node.addedThisSession and not @titleUpdated` in the handler, but it's not a big problem in any case.

I think this is fine in general, and works well.

PR #255, opened 2017-04-05
Fixed undo/redo buttons so they work in standalone mode [#133193161]

dougmartin:
This change disables sending any events to CODAP about undo/redo when we are in standalone mode and instead relies on the Sage undo/redo manager to handle the requsests.

Before this change when standalone mode was detected the Sage undo/redo actions would be skipped and instead CODAP would be informed of the need for undo/redo.  However this effectivly disabled the undo/redo buttons showing in Sage.

A new graph-store variable that tracks if we are running in CODAP (set via codap-connect) is now used to flag when undo/redo events should be sent to CODAP.


ddamelin:
Those under redo buttons need to undo/redo codap graph actions as well in standalone mode. From your description it sounds like only the SageModeler changes will make it to the undo stack.

dougmartin:
@ddamelin Good to know - I haven't really used standalone mode.  I'll have to change the code to allow for that.

dougmartin:
@sfentress / @knowuh - don't review yet - I need to change the code

dougmartin:
I closed this branch and re-opened another one because of a bad merge of the remote branch after a rebase - I had a couple of duplicate commits so I abandoned this branch for another.

PR #258, opened 2017-04-10
Fixed undo/redo buttons so they work in standalone mode [#133193161]

dougmartin:
This change disables sending any events to CODAP about undo/redo when we are in standalone mode and instead relies on the Sage undo/redo manager to handle the requsests.

Before this change when standalone mode was detected the Sage undo/redo actions would be skipped and instead CODAP would be informed of the need for undo/redo. However this effectivly disabled the undo/redo buttons showing in Sage.

A new graph-store variable that tracks if we are running in CODAP (set via codap-connect) is now used to flag when undo/redo events should be sent to CODAP.

NOTE: there needs to be work done in CODAP in the DI API so that CODAP can notify the DI the state of the undo/redo stack.

### DIFF
--- a/src/code/models/codap-connect.coffee
+++ b/src/code/models/codap-connect.coffee
@@ -46,6 +46,8 @@ module.exports = class CodapConnect
         frame   = ret[0]
         context = ret[1]
 
+        @graphStore.setUsingCODAP true
+
         if frame?.values.externalUndoAvailable
           CodapActions.hideUndoRedo()
         else if frame?.values.standaloneUndoModeAvailable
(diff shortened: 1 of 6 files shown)


kswenson:
@dougmartin I'm assigning back to you for final review. I'll be reviewing the CODAP-side changes with Jonathan later today.

dougmartin:
Looks good 👍

dougmartin:
I think it make sense to merge these even without the CODAP support in place - the buttons don't work in standalone mode right now anyway.

kswenson:
Corresponding CODAP PR is [Add canUndo/canRedo values to undo-related plugin API calls](https://github.com/concord-consortium/codap/pull/147).

kswenson:
The corresponding CODAP PR has been merged, so the CODAP support should be in the next CODAP build.

PR #261, opened 2017-04-12
Nodes must have unique names [#134798657]

dougmartin:
Adds a check when either a file is loaded or when a title is edited to require both a unique node name and the non-default node name ("Untitled"). If the name is not unique a number is appended to it starting at either 2 if the existing node name does not end with an integer or the ending integer + 1.

For visual feedback during title editing the input turns red when the name is not unique.

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -154,8 +154,7 @@ GraphStore  = Reflux.createStore
 
   ensureUniqueTitle: (node, newTitle=node.title) ->
     nodes = @getNodes()
-    defaultTitle = tr "~NODE.UNTITLED"
-    if newTitle is defaultTitle or not @isUniqueTitle newTitle, node, nodes
+    if not @isUniqueTitle newTitle, node, nodes
       index = 2
       endsWithNumber = / (\d+)$/
       matches = newTitle.match(endsWithNumber)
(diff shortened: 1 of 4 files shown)


src/code/stores/graph-store.coffee
```diff
@@ -146,9 +146,31 @@ GraphStore  = Reflux.createStore
...
+  isUniqueTitle: (title, skipNode, nodes=@getNodes()) ->
+    nonUniqueNode = (otherNode) ->
+      sameTitle = otherNode.title is title
+      if skipNode then sameTitle and otherNode isnt skipNode else sameTitle
+    not _.find nodes, nonUniqueNode
+
+  ensureUniqueTitle: (node, newTitle=node.title) ->
+    nodes = @getNodes()
+    defaultTitle = tr "~NODE.UNTITLED"
+    if newTitle is defaultTitle or not @isUniqueTitle newTitle, node, nodes
```
kswenson: I was playing around with this before merging it just now and the first node created gets a default name of "Untitled 2", which seems counter-intuitive. My inclination would be to remove the `defaultTitle` check altogether. If the user really wants to name one of their nodes "Untitled" we should let them. They just can't have two of them. 😉

dougmartin: Good catch.  I changed the default index from 1 to 2 at the last minute and didn't retest with a blank canvas.  I'll update and test it without the default check.

dougmartin:
@kswenson You are right it looks much nicer now without the default title change.  I pushed a commit to remove that code.  I'll wait to merge until you have seen it.

dougmartin:
Actually I'm going to go ahead and merge so that I can update PT.

Issue #152, opened 2016-03-29
Uncaught ReferenceError: jsPlumb is not defined


knowuh:
View details in Rollbar: [https://rollbar.com/knowuh/Sage/items/272/](https://rollbar.com/knowuh/Sage/items/272/)

```

ReferenceError: jsPlumb is not defined
  File "https://sage.concord.org/js/app.js", line 27, in Object.window.Sage.initApp
  File "https://sage.concord.org/index.html", line 52, in [anonymous]
```

Issue #151, opened 2016-03-28
TypeError: Cannot read property 'toString' of null


knowuh:
View details in Rollbar: [https://rollbar.com/knowuh/Sage/items/269/](https://rollbar.com/knowuh/Sage/items/269/)

```

TypeError: Cannot read property 'toString' of null
  File "https://sage.concord.org/js/app.js", line 67279, in module.exports.React.createClass.render
  File "https://sage.concord.org/js/globals.js", line 44506, in ReactCompositeComponentMixin._renderValidatedComponentWithoutOwnerOrContext
  File "https://sage.concord.org/js/globals.js", line 44533, in ReactCompositeComponentMixin._renderValidatedComponent
  File "https://sage.concord.org/js/globals.js", line 50992, in wrapper [as _renderValidatedComponent]
  File "https://sage.concord.org/js/globals.js", line 43954, in ReactCompositeComponentMixin.mountComponent
  File "https://sage.concord.org/js/globals.js", line 50992, in wrapper [as mountComponent]
  File "https://sage.concord.org/js/globals.js", line 51697, in Object.ReactReconciler.mountComponent
  File "https://sage.concord.org/js/globals.js", line 50639, in ReactDOMComponent.ReactMultiChild.Mixin._mountChildByNameAtIndex
  File "https://sage.concord.org/js/globals.js", line 50545, in ReactDOMComponent.ReactMultiChild.Mixin._updateChildren
  File "https://sage.concord.org/js/globals.js", line 50490, in ReactDOMComponent.ReactMultiChild.Mixin.updateChildren
```

PR #266, opened 2017-04-19
Synchronize Sage node title/CODAP attribute name changes

kswenson:
- synchronize nodes/attribute IDs and titles/names at initialization time
- serialize synchronized IDs and names/titles for continued synchronization
- call plugin API updateAttributes when node titles are changed [#134408929]
- respond to plugin API updateAttributes notification by changing node title when appropriate [#137900155]

Requires CODAP PR [#151](https://github.com/concord-consortium/codap/pull/151) to provide the necessary changes to the CODAP plugin API. Requires CODAP PR [#152](https://github.com/concord-consortium/codap/pull/152) to support undo/redo of title changes when synchronizing CODAP attribute names.

### DIFF
--- a/src/code/models/codap-connect.coffee
+++ b/src/code/models/codap-connect.coffee
@@ -79,7 +79,7 @@ module.exports = class CodapConnect
       # ret==null is indication of timeout, not an indication that the data set
       # doesn't exist.
       if !ret or ret.success
-        @_initialSyncAttributeProperties ret.values?.collections?[1].attrs
+        @_initialSyncAttributeProperties ret.values?.collections?[1]?.attrs
         @initGameHandler ret
       else
         @_createDataContext()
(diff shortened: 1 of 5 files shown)


src/code/models/codap-connect.coffee
```diff
@@ -77,10 +79,18 @@ module.exports = class CodapConnect
       # ret==null is indication of timeout, not an indication that the data set
       # doesn't exist.
       if !ret or ret.success
+        @_initialSyncAttributeProperties ret.values?.collections?[1].attrs
```
dougmartin: Are you missing an existential operator after the [1] like this:

`@_initialSyncAttributeProperties ret.values?.collections?[1]?.attrs`

src/code/models/codap-connect.coffee
```diff
@@ -172,6 +184,52 @@ module.exports = class CodapConnect
       @codapPhone.call(getListing, doResolve)
       log.info "requested list of attributes"
 
+  _syncAttributeProperties: (attrProps, initialSync) ->
+    nodesToSync = if initialSync \
+                    then _.filter @graphStore.nodeKeys, (node) -> not node.codapID or not node.codapName \
+                    else _.map @graphStore.nodeKeys, (node) -> node
```
dougmartin: This `(node) -> node` looked weird at first but then I realized the _.map callback is `(value, key)` so it works.  It really does look like a noop though.

src/code/models/codap-connect.coffee
kswenson: Good catch; I'll add it.

src/code/models/codap-connect.coffee
kswenson: I added a comment that should make the intent somewhat clearer.

kswenson:
@dougmartin One more commit (fixing undo/redo) to review.

dougmartin:
Looks good 👍

PR #274, opened 2017-04-27
Collector shows multiple images of the assigned icon [#144045227]

kswenson:
- Collectors use a set of five images "randomly" distributed within the node [#144045227]
- in fact the images are in five fixed locations with small random perturbations so that they generally cover the view

Went through several iterations with Dan before settling on this behavior.

### DIFF
--- a/src/code/views/stacked-image-view.coffee
+++ b/src/code/views/stacked-image-view.coffee
@@ -28,6 +28,6 @@ module.exports = React.createClass
       transform: "rotate(#{imgProps.rotation}deg)"
     )
     div { style: { position: "relative", width: "100%", height: "100%" } },
-        @props.imageProps.map((imgProps, index) =>
+        _.map(@props.imageProps, (imgProps, index) =>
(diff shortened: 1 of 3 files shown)


src/code/views/stacked-image-view.coffee
```diff
@@ -0,0 +1,33 @@
...
+    width: "50%"
+
+  render: ->
+    styles = @props.imageProps.map((imgProps) ->
+      top: "#{imgProps.top}%"
+      left: "#{imgProps.left}%"
+      transform: "rotate(#{imgProps.rotation}deg)"
+    )
+    div { style: { position: "relative", width: "100%", height: "100%" } },
+        @props.imageProps.map((imgProps, index) =>
```
dougmartin: This should be changed to _.map as Array.map isn't available in some older browsers.

kswenson: I changed it to `_.map()` for consistency, but `Array.map` was introduced in IE9 and I don't think we support anything older than IE11 at this point.

PR #275, opened 2017-04-28
Fixed 3 bugs seen in search during unrelated PR review

dougmartin:
- changed ImagePicker to use index for key intead of id to make it unique
- fixed @setState on unmounted component in ImageSearchResult
- fixed unselected tab when selecting image from search result [#144495595]

The last bug was introduced in this PR: https://github.com/concord-consortium/building-models/pull/273/

### DIFF
--- a/src/code/views/image-picker-view.coffee
+++ b/src/code/views/image-picker-view.coffee
@@ -46,7 +46,7 @@ module.exports = React.createClass
             label: tr '~PALETTE-INSPECTOR.ADD_IMAGE_SHORT'
           })
         )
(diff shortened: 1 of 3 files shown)


kswenson:
One other thing: check PT. I think Evangeline logged at least one of these bugs during testing yesterday, so you may want to reference the corresponding PT story where one exists.

dougmartin:
@kswenson I updated the PR comment and the commit with the story [#144495595] - https://www.pivotaltracker.com/story/show/144495595

PR #276, opened 2017-04-29
Fix relationship panel tab navigation bug [#144497757]

kswenson:
The `selectedTab` as understood by the TabbedPanelView was getting out of sync with the InspectorPanelStore's notion of the `selectedLink`. The fix implemented here is to configure the TabbedPanelView with an onTabSelected prop which is used to update the `selectedLink`, which eventually propagates back to set the `selectedTab` of the TabbedPanelView.

Note that two recent PRs (#273 & #275) dealt with the TabbedPanelView's setting of its `selectedTab` state. Given that the InspectorPanelStore's `selectedLink` is used to specify the `selectedTab` prop for the TabbedPanelView, it's worth revisiting those fixes to make sure that they're still appropriate in light of these changes.

### DIFF
--- a/src/code/views/relation-inspector-view.coffee
+++ b/src/code/views/relation-inspector-view.coffee
@@ -25,13 +25,16 @@ module.exports = RelationInspectorView = React.createClass
 
     (Tabber.Tab {label: label, component: relationView, defined: isFullyDefined})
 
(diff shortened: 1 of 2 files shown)


dougmartin:
Looks good 👍

PR #278, opened 2017-05-04
Flow ui and simulation changes

dougmartin:
- Updated UI to add transfer variables for flows
  - Added add, subtract and transferred to relationship options [#136570689] [#143266035]
  - Set hidden flow variable name [#143266249]
  - Added fractional options [#143266129]
- Updated simulation to handle flows [#136566677] [#136566739]

NOTE: There are several TODOs marked in the simulation code that relate to questions for @ddamelin .  I wanted to wait to ask the questions when @kswenson was available (at the demo) since he is going to take over this code.  This PR should not be merged until those TODOs are addressed.

Added additional commits to address issues identified after Doug's initial work:
- Improve flow node rendering [#143266035]
- Improve relation inspector wording for transfer-modifier links [#143266249] [#143266129]
- Show relationship symbols on collector links [#136570689]
- Fix bugs in deleting nodes and undoing node/link deletion [#144883527]

@dougmartin I'm assigning to you for review on the work done so far (apparently I can't make you a reviewer on a PR which you created initially), but expect there to be at least one more commit wrapping up the simulation engine changes.

@dougmartin I've added one more commit with the simulation changes, after which I'd like to think this PR is (finally) complete. I'm working on the unit test changes required to accommodate the simulation changes. Please review at your convenience. One more thing: you raised the issue in comments in the code that the use of `previousValue` is not particularly intuitive. How would you feel about replacing it with something like `stableValue` or `settledValue` as a means of making its use somewhat clearer?

@dougmartin I've fixed the unit tests to account for the changes to the simulation engine and added a couple new unit tests for the transfer node functionality. At this point I think it's ready for final review.

@dougmartin Dan has officially cleared the current behavior for merging to master, so if you wouldn't mind reviewing the last couple commits I think we can finally put this PR behind us. 😉 I can merge it if you just review those last few commits. (One of those commits includes changes from your prior review comments.) Thx!


dougmartin:
Looks good so far 👍

src/code/models/simulation.coffee
```diff
@@ -7,92 +7,89 @@ scaleInput = (val, nodeIn, nodeOut) ->
...
+  return null if not inValues?.length
+  return inValues[0] if inValues.length is 1
+
+  if useScaledProduct
+    # scaled product is computed as (n1 * n2 * ...) / 100^(n-1)
+    numerator = _.reduce(inValues, ((prod, value) -> prod * value), 1)
+    denominator = Math.pow(100, inValues.length - 1)
+  else
+    # simple arithmetic mean
+    numerator = _.reduce(inValues, ((sum, value) -> sum + value), 0)
```
dougmartin: Nice use of reduce()

src/code/models/simulation.coffee
```diff
@@ -7,92 +7,89 @@ scaleInput = (val, nodeIn, nodeOut) ->
...
-  value = 0
+  linkScaleFactor = 100
+
+  deltaValue = 0
   for link in links
     {sourceNode, relation, transferNode} = link
     inV = nodeValues[sourceNode.key]
     switch relation.type
       when 'accumulator'
         # TODO: check if outV is really equal to inV here - I'm not clear what outV is in this case'
```
dougmartin: I wrote this TODO because I wasn't sure what outV should be here - @kswenson do you have a handle on that?

src/code/models/simulation.coffee
```diff
@@ -7,92 +7,89 @@ scaleInput = (val, nodeIn, nodeOut) ->
...
+        deltaValue += relation.evaluate(inV, inV, sourceNode.max, @max) / linkScaleFactor
 
       when 'transfer'
-        # TODO: check with Dan if the transfer value really is meant to be a percentage
         transferValue = nodeValues[transferNode.key]
-        transferPercentage = transferValue / Math.max(transferNode.max, transferValue)
-        for modifierLink in transferNode.inLinks('transfer-modifier')
-          transferPercentage *= modifierLink.relation.evaluate(inV, inV, sourceNode.max, transferNode.max)
+        # transfer values are scaled unless they have a modifier
+        if not transferNode.inLinks('transfer-modifier').length
```
dougmartin: What about the modifier link's relation?  Is that now handled in the outer simulation loop?

kswenson: Yes, since `transfer-modifier` links can interact with `range` links on a transfer node, they're handled in the first pass. So `range` and `transfer-modifier` links are handled in the first pass, while `accumulator` and `transfer` links are handled in the second.

src/code/models/simulation.coffee
kswenson: Not yet, that's why I left the `TODO` in place. :wink: I'll take another look, though, and see if I can figure it out.

kswenson:
Thanks for the review. I'm going to wait for Dan to weigh in on whether it behaves the way he expected before merging to master.

ddamelin:
Sorry. I had intended to do more testing this morning, but it got taken up with other issues. I will test as soon as I can, but not until later today, maybe even evening.

src/code/models/simulation.coffee
kswenson: 1. After a fairly thorough search of the code, I don't think it matters, because it doesn't look like it's used anywhere. It's part of the `scope` that is passed to the evaluated formulas, but I didn't find any formulas that reference it.
2. That said, by analogy with its handling in `RangeIntegrationFunction`, I think it's reasonable to assume that the output expected is the previous output of the node, so `@previousValue` or `@initialValue`, which is now captured in the code as `startValue` (as in `RangeIntegrationFunction`).

That's my story and I'm sticking to it. :wink: I've added a commit to that effect.

src/code/data/migrations/19_add_relation_type.coffee
```diff
@@ -0,0 +1,11 @@
+migration =
+  version: "1.18.0"
+  description: "Adds link reasoning"
```
dougmartin: My bad here, should be "Adds link relationship type"

src/code/utils/js-plumb-diagram-toolkit.coffee
```diff
@@ -226,11 +230,16 @@ module.exports = class DiagramToolkit
     if opts.showIndicators? and not opts.showIndicators
       changeIndicator = null
 
+    if opts.isTransfer
+      @kit.importDefaults
+        Connector: ["Flowchart", {}]
+        Anchor:    ["Continuous", { faces:["left","right"] }]
+
```
dougmartin: I wrote this but now I'm wondering if we need an else here to reset the importDefaults to the real defaults?

src/code/data/migrations/19_add_relation_type.coffee
kswenson: I made the change.

src/code/utils/js-plumb-diagram-toolkit.coffee
kswenson: I'm played around with it and it didn't seem to make any difference, but there is another call to `@kit.importDefaults()` at the end of the function which seems to serve the purpose of resetting the defaults, so I added an `Anchor` to that call for good measure.

kswenson:
@dougmartin This branch is now rebased on top of all the other merged PRs. It has been cleared for merge by Dan and merging it is (finally) my highest priority. There have been a couple of commits since your last review that could use a final review.

The one issue that arose when rebasing against the other branch changes is that on this branch you converted the relation `id`s from numeric values to strings. In the Chinese branch I separated the `text` property into a localizable `uiText` property and the grandfathered `text` property (which is now hardcoded to the prior English strings) and which functions more like an ID and is stored in documents. (The only reason I kept the name `text` was to avoid the migration issues.) The question is whether this is a reasonable state of affairs or whether we should consider addressing it, for instance, by combining the `id` and `text` fields, perhaps using the earlier `text` strings as the `id`, eliminate the `text` field, possibly rename `uiText` back to `text` as the localizable string, and deal with the migration issues by migrating `text` to `id` in documents and then serialize `id` rather than `text` moving forward -- or some other combination. Alternatively, we could decide that it's not worth dealing with it right now and kick the can down the road.

kswenson:
Merged after synchronous review of last few commits with Doug. In that meeting we also determined that since the `text` field is sometimes generated dynamically and then is broken up and parsed for its constituent parts, it didn't make sense to try to combine the `id` and `text` fields or do anything at migration time, so we're just leaving things as they are.

PR #281, opened 2017-05-13
Chinese localization

kswenson:
Adds the Chinese localization to Sage. Language displayed to user is currently controlled via URL parameter (e.g. `?lang=zh` for Chinese), but will eventually be controllable via UI and synchronized with CODAP via plugin API.

Improvements that are included in this PR:
- translation module supports multiple languages files
- translation module selects appropriate language file base on URL parameter (for now)
- fix: translation module no longer treats empty translation strings as errors [#145693983]
- fix: eliminate string dependency from `TimeUnits` unit test
- fix: separate use of relationship `uiText` from `text` so serialization works with translated strings [#145488663]
- add line-endings rule to CoffeeLint configuration so that Windows line endings are detected
- fix: "Simulate" text placement [#145414217] item 1
- fix: search button in image search dialog is now localizable [#145414217] item 3
- fix: prev/next "arrows" in search results show up properly [#145414217] item 2

Note that when using Sage with CODAP, saved files store their plugin URLs in the document. Therefore, with current CODAP there's no way for the user to override the language used by a plugin in a saved document. [CODAP PR #160](https://github.com/concord-consortium/codap/pull/160) adds support for the `di-override` URL parameter, which makes it possible to override the plugin URL via CODAP URL parameter.


dougmartin:
Looks good 👍  I'm going to let you merge this one.

PR #282, opened 2017-05-16
Fix bug when editing title after undoing title change [#145244601]

kswenson:
Fix bug when editing title after undoing title change [#145244601] 
Trim node titles to eliminate leading/trailing white space [#144377505]

A recent bug-fix made the <input> DOM element persistent. This also made its contents persistent, which allowed the incorrect contents to be retrieved because the <input> element is uncontrolled. But since we have an onChange handler, we simply make the <input> element controlled at this point.


dougmartin:
Looks good 👍

PR #285, opened 2017-05-18
Fix clickable hidden popup menu [#145575823]

kswenson:


src/code/views/link-relation-view.coffee
```diff
@@ -142,14 +142,14 @@ module.exports = LinkRelationView = React.createClass
       currentOption = scalarSelection.id
 
     # place dropdown but hide it if we haven't selected vector (to keep spacing)
-    style = if @state.selectedVector then {} else {opacity: 0}
+    visClass = if @state.selectedVector then ' visible' else ' hidden'
```
dougmartin: It might be cleaner to move bb-select into the two strings here so you don't have to do the string interpolation on lines 148 and 152 below.

PR #287, opened 2017-05-19
Improve Sage URL query parameter handling

kswenson:
- fix bug in which a URL like "http://sage.concord.org/?lang=zh" will reload the page infinitely adding "sage.html" to the URL each time
- preserve query parameters when redirecting
- pass query parameters through CFM to embedded app
- fix typo "SageCloundFileManager" -> "SageCloudFileManager"

@knowuh I'm assigning this one to you for review since you seem to have written some of the code being modified here (and I've been hitting up Doug pretty hard lately).


knowuh:
@kswenson looks good.   👍

PR #289, opened 2017-05-24
Only "ensure" unique node names upon completion of user edit

kswenson:
Only "ensure" unique node names upon completion of user edit [#145949051]
- don't try to replace what the user is typing while he/she is typing it


dougmartin:
Looks good 👍

PR #290, opened 2017-05-24
Changing # steps in simulation triggers redraw of mini-graphs

kswenson:
- while in the neighborhood, changing the "Limit values to min/max range" setting also triggers redraw


dougmartin:
Looks good 👍

PR #296, opened 2017-05-25
Regular nodes in closed loops with collectors should not have sliders

kswenson:
Regular nodes in closed loops with collectors should not have sliders [#145981913]
- extends existing closed loop detection mechanism to check for collectors in the closed loop

### DIFF
--- a/src/code/models/node.coffee
+++ b/src/code/models/node.coffee
@@ -54,6 +54,7 @@ module.exports = class Node extends GraphPrimitive
     @color ?= Colors.choices[0].value
 
     @isInCycle = false      # we always initalize with no links, so we can't be in cycle
+    @hasAccumulatorAncestor = false
 
     @_collectorImageProps = null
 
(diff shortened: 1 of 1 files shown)


src/code/models/node.coffee
```diff
@@ -131,20 +132,23 @@ module.exports = class Node extends GraphPrimitive
   checkIsInIndependentCycle: ->
     visitedNodes = []
     original = this
-    isOwnGrandpa = false
```
sfentress: I see we're not fans of the song "I'm My Own Grandpa". No matter. 😜

kswenson:
Eliminating `hasAccumulatorAncestor` was an improvement as you suggested, so I've made the corresponding changes.

src/code/models/node.coffee
kswenson: The main reason I eliminated it was for consistency. I didn't want to add `hasAccumulatorGrandpa` and having two variables that referred to the same concept using different terms seemed potentially confusing. Now that we've eliminated `hasAccumulatorAncestor` (at your suggestion), however, the consistency argument no longer applies so I've restore `isOwnGrandpa`.

sfentress: Ha! I really was kidding. Clarity is most important. But either is good.

PR #298, opened 2017-05-26
Add "Cannot go below zero" collector setting [#145734901]

kswenson:
- user setting defaults to true
- internal setting (`allowNegativeValues`) defaults to false for backward compatibility
- when set collector value is limited to non-negative values
- when set on source node limits transfer values so as not to overdraw source
- setting is disabled/checked when global "Limit values to min/max range" setting is enabled
- tooltips provide more information (on desktop) on the various states
- added four new strings to strings files; new strings have # TRANSLATE comment in Chinese strings file


src/code/models/simulation.coffee
```diff
@@ -34,6 +34,19 @@ combineInputs = (inValues, useScaledProduct) ->
 
   numerator / denominator
 
+getTransferLimit = (transferNode) ->
+  return 0 if not transferNode?.transferLink?.sourceNode?
+  if transferNode.transferLink.sourceNode.previousValue? \
+    then transferNode.transferLink.sourceNode.previousValue \
+    else transferNode.transferLink.sourceNode.initialValue
+
```
kswenson: Nice! I made a slightly different change based on your suggestion.

PR #299, opened 2017-05-27
Synchronize use of duration throughout system

kswenson:
- fix one-step recording bug (should only record one data point)
- fix mini-graphs bug in which the first point(s) weren't plotted [#146030015]
- fix bug in which cached mini-graph data could grow unbounded [#146183765]
- adjust simulation unit tests to test initial values

One of the changes I made to the simulation loop while making the recent set of improvements to support flow nodes and collectors better was to add one step to the simulation loop to account for the initial conditions, i.e. if the user asks for ten steps, eleven data points should be generated -- the initial conditions plus the results of the first ten steps. This made sense in that context, but had some less-than-desirable results in other contexts, notably that one-step recording would now record two steps and mini-graphs were not displaying those initial conditions.

The approach taken here is to eliminate the additional step from the `Simulation` model -- when you ask for a specific duration, that's the number of steps you get -- and put the responsibility on the client to determine when to ask for the additional step. This puts the nexus of responsibility in the `SimulationStore` to determine the appropriate duration. Clients like the mini-graphs can use methods like `SimulationStore.simulationDuration()` to determine the total length of the simulation, which allows the mini-graphs to plot all of the data without missing the first couple points.

While investigating the mini-graph issues described above I discovered the fact that while dragging sliders to see the effect on the mini-graphs, the amount of data cached for use by the mini-graphs continued to grow until nodes were caching thousands of points, which would then be serialized with the document. Curious, I looked back through the git history where I discovered that originally, running the simulation triggered the clearing of mini-graph caches. When the one-step recording feature was added, this automatic clearing of the mini-graph caches was removed because in that case each additional point should be added to the mini-graph cumulatively. But no alternate mechanism for clearing the mini-graph caches was ever put in place. The solution I'm using here is to trim the size of the cache on each simulation run, rather than clearing it entirely.

Finally, moving the specification of the additional step out of the `Simulation` model required changes to the unit tests, which actually make the tests more clear because the initial conditions can now be tested whereas they could get hidden by the implicit additional step.

@knowuh I'm tapping you for review since you were part of the mini-graph and one-step recording work.

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -44,7 +44,6 @@ GraphStore  = Reflux.createStore
     @updateListeners()
 
   _trimSimulation: ->
-    #settings = SimulationStore.store.serialize()
     for node in @getNodes()
       # leaving some excess data reduces flicker during rapid changes
       excessFrames = node.frames.length - 2 * SimulationStore.store.simulationDuration()
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -43,12 +43,20 @@ GraphStore  = Reflux.createStore
       node.frames = []
     @updateListeners()
 
+  _trimSimulation: ->
+    #settings = SimulationStore.store.serialize()
+    for node in @getNodes()
+      # leaving some excess data reduces flicker during rapid changes
+      excessFrames = node.frames.length - 2 * SimulationStore.store.simulationDuration()
+      if excessFrames > 0
+        node.frames.splice(0, excessFrames)
+    return  # prevent unused default return value
+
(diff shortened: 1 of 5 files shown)


src/code/stores/graph-store.coffee
```diff
@@ -43,12 +43,20 @@ GraphStore  = Reflux.createStore
       node.frames = []
     @updateListeners()
 
+  _trimSimulation: ->
+    #settings = SimulationStore.store.serialize()
```
knowuh: rm this comment?

kswenson: Good catch. I'll remove it before merging.

PR #303, opened 2017-05-31
Change polarity of collector range checkbox label [#146383493]

kswenson:
- "Cannot go below zero" -> "Can go below zero"


src/code/views/node-value-inspector-view.coffee
```diff
@@ -46,7 +46,7 @@ module.exports = React.createClass
     SimulationStore.actions.toggledCollectorTo value
 
   updateNegativeValuesAllowed: (evt) ->
-    value = !evt.target.checked   # checkbox is reverse polarity from property
+    value = !!evt.target.checked
```
dougmartin: Did you mean to use `!!``?  That just converts a truthy/falsy value to a boolean and checked is already a boolean.

PR #305, opened 2017-06-02
Final tweaks

kswenson:
- Fix bug found by Dan in which flows from an empty source collector could increase destination collector in some circumstances
- Adjust endpoints along bottom of flow node


ddamelin:
👍

PR #311, opened 2017-09-07
Support POEditor; add Turkish & Hebrew translations

kswenson:
- Master English strings file is now `src/code/utils/lang/en-US-master.json`, which is a strict JSON file except that comments are allowed. (Comments are stripped before use where appropriate.) In other words, code changes should be made to `src/code/utils/lang/en-US-master.json`, _not_ `src/code/utils/lang/en-US.json` (which is now an output file generated by script).
- Changes to master English strings file can be pushed to POEditor with `npm run poeditor:push`. (POEditor API key must be specified as command-line argument or set as environment variable. It is too sensitive to be committed to Github.)
- Updated translations can be pulled from POEditor with `npm run poeditor:pull`. (POEditor API key must be specified as command-line argument or set as environment variable. It is too sensitive to be committed to Github.) Note that the pull command also copies the master English strings file into its JavaScript form in the project, so it should be used to update the English as well.
- Testable at http://sage.concord.org/branch/translate/sage.html?lang=tr (Turkish) and http://sage.concord.org/branch/translate/sage.html?lang=he (Hebrew).

@dougmartin Don't merge until we've gotten more feedback about the overall approach. [old]
@sfentress I think this has been discussed enough that it's ready to be merged. The corresponding CODAP and CFM PRs have already been merged. I rebased this to master and did some manual merging of your recent strings and dependency changes. You should at least review those changes to make sure I captured everything correctly. You may choose to review the rest of the PR or not as you see fit. As your recent changes suggest, the longer this sits out there the more likely we are to encounter merge conflicts moving forward.


package.json
```diff
@@ -11,6 +11,10 @@
     "build": "gulp build-all",
     "deploy:dev": "./build.sh",
     "lint": "gulp lint",
+    "poeditor:pull:comment": "echo Usage: `npm run poeditor:pull -- -a <your_poeditor_api_key>`",
```
dougmartin: What about using `:usage` instead of `:comment`?

dougmartin: Or `:help`

PR #314, opened 2017-09-27
Complexity levels

sfentress:
Hi @knowuh, this is a bigger PR. If you could look at it any time in the next few days that would be great.

This adds more settings for the "complexity" of a diagram, beyond the original "diagram-only tools" checkbox that existed before.

You can now set the complexity to four levels, "diagram-only", "basic relations", "expanded relations" and "collectors." By default a model starts at "basic", unless (1) it loads data that had the old `diagramOnly: true` setting, or (2) you launch with the hash param `#simplified=true` (matching the old feature).

Lower complexity levels prevent the user from accessing certain tools, like the "varies" relationship, all the scalar relationships, and turning nodes into collectors.

The graph is scanned to note the minimum complexity it supports, and prevents the user from lowering the complexity of a diagram if there are features present of a higher complexity. (If the user removes those features, they can subsequently lower the complexity.)

See PT story: https://www.pivotaltracker.com/story/show/151201919

Built branch: http://sage.concord.org/branch/complexity-levels/sage.html

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -14,6 +14,7 @@ GraphActions        = require "../actions/graph-actions"
 CodapActions        = require '../actions/codap-actions'
 InspectorPanelStore = require "../stores/inspector-panel-store"
 CodapConnect        = require '../models/codap-connect'
+RelationFactory     = require "../models/relation-factory"
 DEFAULT_CONTEXT_NAME = 'building-models'
 
 GraphStore  = Reflux.createStore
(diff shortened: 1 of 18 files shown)


src/code/stores/graph-store.coffee
```diff
@@ -516,6 +517,41 @@ GraphStore  = Reflux.createStore
...
+    links = @getLinks()
+    _.each links, (link) ->
+      return unless (source = link.sourceNode) and (target = link.targetNode)
+
+      if source.isAccumulator or target.isAccumulator
+        # we know we have to be the highest complexity
+        minComplexity = 3
+      else if link.relation?.formula
+        # we know we'll be at least 1 or 2
+        linkComplexity
```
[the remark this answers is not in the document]
sfentress: D'oh, old habits die hard. It was the equivalent of putting `var linkComplexity` before we set it in the if-block, but I forget that Coffeescript does that for you.

sfentress:
@knowuh I made another change, which was to replace the complexity values with ints rather than strings. You didn't mention it, but I thought all the comparisons were verbose and ugly, and this helped to streamline it. The fact that "min complexity" is a thing makes it seem like it should explicitly be numeric. Thoughts?

test/migrations-test.coffee
```diff
@@ -138,4 +138,4 @@ describe "Migrations",  ->
 
     describe "v-1.19.0 changes", ->
       it "should have complexity setting", ->
-        @result.settings.complexity.should.equal "basic"
+        @result.settings.complexity.should.equal 1
```
knowuh: Everywhere else you have used the more literate style eg `AppSettingsStore.Complexity.basic`.  Is there a reason not to do that here?

src/code/views/simulation-inspector-view.coffee
```diff
@@ -31,25 +31,15 @@ module.exports = React.createClass
...
     diagramOnly = @state.complexity is Complexity.diagramOnly
     if diagramOnly then runPanelClasses += " collapsed"
 
     minComplexity = GraphStore.getMinimumComplexity()
-    diagramOnlyDisabled = @isDisabled(Complexity.diagramOnly, minComplexity)
-    basicDisabled = @isDisabled(Complexity.basic, minComplexity)
-    expandedDisabled = @isDisabled(Complexity.expanded, minComplexity)
+    diagramOnlyDisabled = minComplexity > Complexity.diagramOnly
+    basicDisabled = minComplexity > Complexity.basic
+    expandedDisabled = minComplexity > Complexity.expanded
```
knowuh: This *is* much cleaner.  👍

test/migrations-test.coffee
sfentress: Elsewhere, everything can reference the same the same value for `AppSettingsStore.Complexity.basic` and stay internally consistent, so if we change it under the hood, nothing at all changes (well, except for the numeric comparisons, now). However, the saved data exists independently of that, and will always be `1` once set that way. It seemed that making this explicit would help us avoid errors if, say, we change the value of `AppSettingsStore.Complexity.basic` and just forget that old data will still be using `1`. Does that make sense?

knowuh: If we keep it `AppSettingsStore.Complexity.basic` a failing test will remind us to handle a migration to update the number?  Otherwise, I see your reasoning, and  I am quite fine with ether.

sfentress: I guess it's a little odd, because the migration itself uses the property key. But I just feel like the migrations tests should test the raw data, not our interpretations of the data, because that will be the literal result of the migration.

PR #315, opened 2017-09-29
Set collector initial value

sfentress:
@kswenson since you were most recently working in the simulation code, I wonder if you could take a look at this.

This implements this story: https://www.pivotaltracker.com/story/show/151201810

Non-collector variables can link to collectors with a new relationship type, "set initial value." The model should run exactly the same as if the collector's own initial value had been set by hand.

The simulation changes ended up being pretty simple once I realized I should just set the initial values before running through the steps.

The model can be run here: http://sage.concord.org/branch/set-collector-init-value/sage.html
In CODAP: http://codap.concord.org/releases/latest/static/dg/en/cert/index.html?di=http://sage.concord.org/branch/set-collector-init-value/sage.html

(en-US-master.json updated, en-US.json updated so that the branch deploy would have strings, and will be overwritten before merging by POEditor push/pull.)

### DIFF
--- a/src/code/models/simulation.coffee
+++ b/src/code/models/simulation.coffee
@@ -82,6 +82,19 @@ RangeIntegrationFunction = (incrementAccumulators) ->
 
   value
 
+# Sets the value of node.initialValue before the simulations starts. If there
+# are inbound `initial-value` links, we request the initial values of the
+# source nodes (no calculations needed) and average them.
+SetInitialAccumulatorValueFunction = () ->
+  initialValueLinks = @inLinks('initial-value')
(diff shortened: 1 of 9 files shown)


src/code/models/simulation.coffee
```diff
@@ -217,6 +231,10 @@ module.exports = class Simulation
     nodeValues = {}
     collectorNodes = _.filter @nodes, (node) -> node.isAccumulator
 
+    # before the first step, set the initial values of all aqccumulators,
```
kswenson: typo: "accumulators"

PR #320, opened 2018-01-08
Changes to flow-modifying links.

knowuh:
This PR changes the way a source (and now a target) node can change their flow relationship.

The goal is to reduce the number of flow modifiers and allow both the source and target nodes to modify the flow.

* There are now just two flow-modifiers: proportional and inverse proportional flow.
* Target nodes can now affect the flow.

There have been modifications to the text and language files.  The text has to change significantly when the target node is changing the flow.

Also as part of this PR, the default MAX for collectors is 1000, and the default simulation length is 20 steps.  These changes were made so that the effect of flow-modifiers could be easily seen using system defaults in the mini-graphs.

Tests which that involved the MAX of collectors or the formulas for flow-modifications had to be adjusted.

### DIFF
--- a/src/code/models/node.coffee
+++ b/src/code/models/node.coffee
@@ -6,6 +6,7 @@ urlParams = require '../utils/url-params'
 
 SEMIQUANT_MIN = 0
 SEMIQUANT_MAX = 100
+SEMIQUANT_ACCUMULATOR_MAX = 1000
 
 # Enable ES5 getters and setters
 Function::property = (prop, desc) ->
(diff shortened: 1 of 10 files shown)


sfentress:
Quick feature-level question: is it correct/expected that if a user does not add a flow-modifier link to a transfer relationship, that the transfer does nothing? This seems to make it harder to just set up a simple flow, and makes a straight-forward linear transfer impossible.

Old Sage:

screen shot 2018-01-09 at 3 47 11 pm

New:

screen shot 2018-01-09 at 3 47 22 pm

sfentress:
Ah, I see what's happening, the default transfer is still occurring, but something to do with the new values (maybe the new max) means that the decrease isn't visible in the graph after 50 steps. Setting the simulation time to 200 makes the decrease visible.

I wonder whether the default transfer should be increased, to meet expectations?

Actually, there may be a separate bug -- if I have a transfer set up with no modifications, and the default initial values, I see no visible decrease. If I switch Node A to defining with numbers, I *do* see a decrease. Is the scaling different from quant and semi-quant?

Semi-quant:

screen shot 2018-01-09 at 4 02 13 pm

Quant:

screen shot 2018-01-09 at 4 02 24 pm

ddamelin:
@sfentress You are seeing this behavior because @knowuh has implemented half of the changes we discussed in the algorithm. The change you are seeing here is the default range for a collector is now 0 to 1000, but we still have in place the special case where a source with no links to or form it besides the transfer divides the flow value by 100. Eliminating that special case will be in another (hopefully soon to arrive) PR.

sfentress:
The code all looks good. :+1:

PR #321, opened 2018-01-10
Several small collector / flow adjustments

knowuh:
@sfentress would you mind looking at this PR?  

It's smaller than the last, and related. Hopefully, it's a quick read.

Summary of Changes:
* The calculation for a flow by default should use `limiting factor` (aka `product`)
* Simulation reruns when the transfer's `combine method` is changed, so graphs update.
* Remove a special case of dividing a collectors value by 100 if there is no input. (This is the one you and Dan were discussing in the last PR)
* Immediately enable `display mini-graphs` if we are running a simulation with collectors. 

Some tests needed to be adjusted to handle the divide by 100 change.

Related PT Stories:
https://www.pivotaltracker.com/story/show/153639811
https://www.pivotaltracker.com/story/show/153639752
https://www.pivotaltracker.com/story/show/154029978
https://www.pivotaltracker.com/story/show/153639400
https://www.pivotaltracker.com/story/show/153639740
https://www.pivotaltracker.com/story/show/153639575

### DIFF
--- a/src/code/views/simulation-run-panel-view.coffee
+++ b/src/code/views/simulation-run-panel-view.coffee
@@ -23,8 +23,10 @@ module.exports = React.createClass
       SimulationStore.actions.collapseSimulationPanel()
     else
       SimulationStore.actions.expandSimulationPanel()
-      if ! @state.showingMinigraphs
-        AppSettingsStore.actions.showMinigraphs true
+      # -- TBD: There was discussion about automatically showing
+      # -- MiniGraphs when this panel is opened  …  NP 2018-01
+      # if ! @state.showingMinigraphs
+      #   AppSettingsStore.actions.showMinigraphs true
 
   renderToggleButton: ->
     iconClass = if @state.simulationPanelExpanded then "inspectorArrow-collapse" else "inspectorArrow-expand"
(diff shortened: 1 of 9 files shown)


knowuh:
OK @sfentress  -- sorry fo the confusion / delay.  This is probably ready to be reviewed again.  Despite the large number of commits, it shouldn't be a huge review.

src/code/views/simulation-run-panel-view.coffee
```diff
@@ -21,6 +23,10 @@ module.exports = React.createClass
       SimulationStore.actions.collapseSimulationPanel()
     else
       SimulationStore.actions.expandSimulationPanel()
+      # -- TBD: There was discussion about automatically showing
+      # -- MiniGraphs when this panel is opened  …  NP 2018-01
+      # if ! @state.showingMinigraphs
+      #   AppSettingsStore.actions.showMinigraphs true
```
sfentress: This comment is no longer applicable, because this feature is implemented elsewhere, right?

ddamelin:
Yes. When they switch a regular variable to a collector we should change the current value proportionally (and vice versa if they switch back from being a collector with a cap of the max of the range). Perhaps that needs to be another story. (So if the value is 50 for a range of 0 to 100 it becomes 500 for a collector with a range of 0 to 1000)

knowuh:
It would be great to put in a new PT story for changing the value proportionally, because I have moved on to a different project.

PR #326, opened 2018-04-02
Add simulation type

sfentress:
Hi Noah, this PR is for 4 stories:

* Segment the levels of complexity: https://www.pivotaltracker.com/story/show/153599165
* When we switch to time-based, use time-based experiments even if there are no collectors: https://www.pivotaltracker.com/story/show/156410734
* Change static equilibrium models to have bar-graph-like mini graphs: https://www.pivotaltracker.com/story/show/156374585
* Move sliders to left side of variables: https://www.pivotaltracker.com/story/show/156374559

The end-result for the user is that, if you are modeling in static mode, you cannot use collectors and you should see bar graphs for minigraphs. If you are modeling in time mode, you may have collectors, and your experiments will be time-based regardless of whether you have collectors.

This adds a new property in the saved data, `simulationType`, and a data migration to go from the old `complexity` to the new `complexity, simulationType` tuple.


src/code/stores/graph-store.coffee
```diff
@@ -529,32 +529,43 @@ GraphStore  = Reflux.createStore
...
-    _.each links, (link) ->
-      return unless (source = link.sourceNode) and (target = link.targetNode)
+    for link in links
+      continue unless (source = link.sourceNode) and (target = link.targetNode)
 
       if source.isAccumulator or target.isAccumulator
-        # we know we have to be the highest complexity
-        minComplexity = 3
+        # we know we have to be time-based
+        return 2
```
knowuh: Minor nitpick: I would rather read `return AppSettingsStore.store.SimulationType.time` used throughout.

knowuh: (or a shorter local name)

sfentress: :+1: You're right, fixed.

PR #328, opened 2018-07-18
Stop UI elements appearing in Diagram-Only mode

dstrawberrygirl:
Double-clicking a link shouldn't open the relationship window, neither should double-clicking a node when we're in that mode. This fix blocks those actions specifically, while leaving the menu functionality intact to enable the user to change to a different mode if they wish.

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -391,7 +391,7 @@ GraphStore  = Reflux.createStore
 
     if isDoubleClick
       @selectionManager.selectNodeForInspection(link.targetNode)
-      if AppSettingsStore.store.settings.simulationType != 0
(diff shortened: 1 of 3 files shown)


src/code/stores/graph-store.coffee
```diff
@@ -391,7 +391,8 @@ GraphStore  = Reflux.createStore
 
     if isDoubleClick
       @selectionManager.selectNodeForInspection(link.targetNode)
-      InspectorPanelStore.actions.openInspectorPanel('relations', {link: link})
+      if AppSettingsStore.store.settings.simulationType != 0
```
kswenson: These tests would be clearer if expressed in terms of a `DIAGRAM_ONLY` constant rather than 0.

PR #330, opened 2018-07-26
DAT-Cross UI changes

dstrawberrygirl:
For the DAT-Cross project:

- Added `lockdown` url hash parameter to limit the ability of users to delete nodes. lockdown=true means users can only remove relationships, not nodes themselves. The assumption is that these users will have models loaded on initialization and will not need a node palette.

- Added `hide` url hash parameter that can remove any of the following pieces from the layout:
```
globalNav
actionBar
inspectorPanel
nodePalette
```

Additional style settings have been added to ensure a viable layout when any of these elements are removed from the UI. In most cases, `display:none` is sufficient to visibly remove them without significant code rework to handle them not being present.

- Added `delete` button to the action bar next to `undo/redo` buttons. This only appears if anything is selected, and can be clicked to delete the selected nodes and relationships. In `lockdown` mode, this only appears if one or more links are selected and will only remove selected relationships

A full test of the end-user experience for DAT-Cross students would require adding the following parameters to the url:
`#hide=inspectorPanel,nodePalette&simplified=true&lockdown=true`

Until loading can be added from the iFramed interactive, the nodePalette should probably be displayed so that the simulation can be tested:
`#hide=inspectorPanel&simplified=true&lockdown=true`

### DIFF
--- a/src/code/views/app-view.coffee
+++ b/src/code/views/app-view.coffee
@@ -50,12 +50,14 @@ module.exports = React.createClass
             username: @state.username
             graphStore: @props.graphStore
             GraphStore: @GraphStore
+            display: AppSettingsStore.store.settings.uiElements.globalNav
           )
         (div {className: 'action-bar'},
           (NodeWell {
             palette: @state.palette
             toggleImageBrowser: @toggleImageBrowser
-            graphStore: @props.graphStore
+            graphStore: @props.graphStore,
+            display: AppSettingsStore.store.settings.uiElements.actionBar
           })
           (DocumentActions
             graphStore: @props.graphStore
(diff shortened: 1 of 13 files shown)


src/code/views/app-view.coffee
```diff
@@ -44,26 +44,28 @@ module.exports = React.createClass
...
+        if not @state.iframed && AppSettingsStore.store.settings.uiElements.globalNav != false
           (GlobalNav
             filename: @state.filename
             username: @state.username
             graphStore: @props.graphStore
             GraphStore: @GraphStore
+            display: AppSettingsStore.store.settings.uiElements.globalNav
           )
-        (div {className: 'action-bar'},
+        (div {className: if AppSettingsStore.store.settings.uiElements.actionBar == false then 'action-bar hidden' else if AppSettingsStore.store.settings.uiElements.globalNav is false then 'action-bar small' else 'action-bar'},
```
knowuh: &^&^$# CoffeeScript! 

I find this line hard to read.  Could you extract a variable here `actionBarStyle` on line 45 or so, and use `if` `else` indenting?   

I have seen people use this style

knowuh:
@dstrawberrygirl  looks great!

knowuh:
@dstrawberrygirl  -- Just a note to remind you this PR is ready for you to merge.

PR #333, opened 2018-07-26
Graphs should update when Combine Variables Using menu changes.

knowuh:
This had stopped working, because we used to save `combineMethod` as a property on flow links. Flow links are no longer a thing, so now we store this as a property on the Nodes.  We have to add `combineMethod` to the serialization hash, so that the graphView knows when it needs to re-run the simulation.

Story:
Graphs should update when Combine Variables Using menu changes.
[#154029978]
https://www.pivotaltracker.com/story/show/154029978

### DIFF
--- a/src/code/stores/graph-store.coffee
+++ b/src/code/stores/graph-store.coffee
@@ -514,7 +514,6 @@ GraphStore  = Reflux.createStore
       linkDescription += "#{source.x},#{source.y};"
       linkDescription += link.relation.formula + ";"
       linkDescription += "#{target.x},#{target.y}|"
-
       if link.relation.isDefined
         isCappedAccumulator = source.isAccumulator and not source.allowNegativeValues
         capValue = if isCappedAccumulator then ':cap' else ''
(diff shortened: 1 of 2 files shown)


src/code/stores/graph-store.coffee
```diff
@@ -524,7 +523,7 @@ GraphStore  = Reflux.createStore
           transfer = link.transferNode
           modelDescription += "#{transfer.key}:#{transfer.initialValue}:#{transfer.combineMethod};" if transfer
         modelDescription += "#{target.key}#{if target.isAccumulator then ':'+(target.value ? target.initialValue) else ''}|"
-
+        modelDescription += target.combineMethod
```
kswenson: It seems to be customary to put a delimiter at the end of each chunk. Otherwise, the `combineMethod` would be concatenated with the subsequent node count.

knowuh: Good catch!  -- Addressed in 086e106

knowuh:
Thanks for reviewing @kswenson  -- I added string delimiters in 72490a0

knowuh:
@dstrawberrygirl  thanks for noticing that.   Fixed in 914d732

PR #334, opened 2018-07-26
159348401 fix tests not running

knowuh:
The mocha tests for building models weren't  running for about 4 months, because of a `describe.only` stanza that was accidentally left in place …

So this PR fixes that, and also cleans up some CodapConnect stubbing that was happening.

The changes are not actually as extensive as they look, but because of the how git-diffing worked in this instance, it seems like there are a lot of changes.

The main thing is that lots of ` @sandbox.stub CodapConnect, "instance", ->`  were replaced with a helper method `CodapHelper.Stub()`  (and `CodapHelper.UnStub()` ).   

This helper was extracted because as we start expecting more messages to be sent to CodapConnect, 
 (such as `_createMissingDataAttributes` it would be nice to only add them once.

@kswenson I have been bugging @dstrawberrygirl for many PR reviews.   If you wanted to peek at this, It might save her eyeballs. 👀 

Thanks!

### DIFF
--- a/test/codap-helper.coffee
+++ b/test/codap-helper.coffee
@@ -1,0 +1,17 @@
+global.window = { location: '' }
+global._      = require 'lodash'
+global.log    = require 'loglevel'
+global.Reflux = require 'reflux'
+
+Sinon          = require('sinon')
+requireModel = (name) -> require "#{__dirname}/../src/code/models/#{name}"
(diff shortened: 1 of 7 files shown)


knowuh:
BTW @dstrawberrygirl --- this is probably why you were so frustrated writing those tests -- they weren't running because of the `.only` thing.  :(

PR #336, opened 2018-07-30
Delete button UI

dstrawberrygirl:
Limiting the appearance of the delete button (next to undo/redo when not inside of CODAP) to only show if the inspectorPanel is hidden and if the user is on a touch device.

### DIFF
--- a/src/code/mixins/app-view.coffee
+++ b/src/code/mixins/app-view.coffee
@@ -20,6 +20,14 @@ module.exports =
   componentDidUpdate: ->
     log.info 'Did Update: AppView'
 
+  addTouchDeviceHandler: (add) ->
+    if add
+      $(window).on 'touchstart', (e) ->
+        AppSettingsStore.actions.setTouchDevice true
+        $(window).off 'touchstart'
+    else
+      $(window).off 'touchstart'
+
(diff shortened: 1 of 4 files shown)


src/code/mixins/app-view.coffee
```diff
@@ -20,6 +20,14 @@ module.exports =
   componentDidUpdate: ->
     log.info 'Did Update: AppView'
 
+  addTouchDeviceHandler: (add) ->
+    if add
+      $(window).on 'touchstart', (e) ->
+        AppSettingsStore.actions.setTouchDevice true
+        $(window).off 'touchstart'
+    else
+      $(window).off 'touchstart'
```
kswenson: As mentioned in the slack thread, this has the effect of always showing the non-touch UI initially and then suddenly switching to the touch UI when the first touch event occurs. If you make an initial assumption based on a static determination (e.g. that iOS/Android browsers are probably touch-centric) but are then able to adjust based on user interaction, you can reduce (but not eliminate) the likelihood that the UI changes unexpectedly.

dstrawberrygirl: Thanks for the feedback - I've added a simple userAgent test in line with this suggestion.

PR #337, opened 2018-08-02
DAT Cross LARA API support initial draft

dstrawberrygirl:
Keeping the code as simple as possible for the C4 integration requirements for Sage, this PR adds basic listeners and calls via IFramePhone for communication with the parent. The API test page we have at http://concord-consortium.github.io/lara-interactive-api/ does not support `getInteractiveState` / `interactiveState` which I believe are the current calls for requesting and responding with the current state of the interactive, so I had to craft a test page, `test-iframephone-parent.html`,  to verify those methods were working, as well as adding listeners for response and log messages. 

Also includes the hash-or-querystring fix for url parameters - if we're outside of an IFrame, we can use hash params, but if we're inside we need query parameters for the hosted IFramed Sage instance to show or hide UI pieces.

Added a test html page to simulate the LMS on the other side that embeds Sage in an IFrame and communicates to and from the page with log messages for debugging, and to assist with external team development by providing a working sample environment.

A rough sketch of the communication back and forth can be found here: http://short.concord.org/6o


knowuh:
Hey @dougmartin 

Dave and I reviewed this, and it looks pretty good.  Would you mind having a quick look?  It should be merged into master before we get much further with Sage.

Thanks 🙏

PR #340, opened 2018-10-12
Changed live-server to use https [#161183550]

dougmartin:
Now that CODAP forces https the local Sage development environment needs to also run under https for the iframe to work.

Also added missing live-server devDependency and change the hostname it uses from 127.0.0.1 to localhost so that Google Drive authentication works.


kswenson:
The other way to deal with this issue is to run the development version of the plugin in a development version of CODAP, which requires setting up the CODAP/SproutCore build environment and using `sc-server`.

dougmartin:
You just need to click the "Advanced" word and allow the site.

dougmartin:
We need https in dev mode because codap forces https so inorder to iframe a local Sage it has to be running under https.

PR #341, opened 2018-10-12
Fixed nextID generation after file is loaded [#160661756]

dougmartin:
Fixes a bug where new nodes sometimes could not be dropped because the node id already existed in the loaded file.

This change sets the counters used in the nextID generation to be the max  + 1 of each node type.


knowuh:
Nice catch.  

This looks good. It makes me wonder how opening *any* saved document worked before this commit.

Would it be possible to create a failing model test in prior commit?  It would just have to load a serialized diagram, and assert that adding a new node would created a duplicate id.

dougmartin:
Good point on the test, will do.

PR #342, opened 2018-10-15
Change node title input to not update on each keypress [#160668583]

dougmartin:
Now the node title only updates after the input is blurred, either by clicking away from the input or hitting the enter key.

The escape key now cancels the node title change, restoring it to the starting title value before the input is shown.

### DIFF
--- a/src/code/views/node-view.coffee
+++ b/src/code/views/node-view.coffee
@@ -17,29 +17,32 @@ NodeTitle = React.createFactory React.createClass
   mixins: [require '../mixins/node-title']
 
   getInitialState: ->
-    isUniqueTitle: @isUniqueTitle @props.node.title
+    @getStateFromProps @props
 
-  isUniqueTitle: (title) ->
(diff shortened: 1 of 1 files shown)


src/code/views/node-view.coffee
```diff
@@ -17,50 +17,65 @@ NodeTitle = React.createFactory React.createClass
...
     if @props.isEditing
       @inputElm().off()
 
   componentWillUpdate: (nextProps) ->
-    # mark the title as updated even if no change was made when it leaves edit mode
-    @titleUpdated = true if @props.isEditing and not nextProps.isEditing
+    if @props.isEditing and not nextProps.isEditing
+      # mark the title as updated even if no change was made when it leaves edit mode
+      @titleUpdated = true
+    else if not @props.isEditing and nextProps.isEditing
```
knowuh: This one line was a little tricky to wrap my head around, but otherwise this whole PR is good.

Thanks!

knowuh:
Going to have you merge this yourself -- looks fine.

PR #343, opened 2018-10-15
159706721 migration incorrectly setting simulation type

DALove1025:
[NP/DL]  Created a test and added a migration step that corrects the simuationType based on if the model contains one or more collector nodes.

PT User Story: https://www.pivotaltracker.com/story/show/159706721

"When an older model contains only flows and collectors simulation settings aren't set correctly when opened by the new version of SageModeler"

This story contains a link to the CODAP document that exhibits the buggy behavior.


src/code/data/migrations/24_fix_simulation_type.coffee
```diff
@@ -0,0 +1,18 @@
...
+migration =
+  version: "1.23.0"
+  description: "Fix simulation type - 1.21.0 to 1.22.0 sim-type broken"
+  date: "2018-10-15"
+
+  doUpdate: (data) ->
+    data.settings ?= {}
+
+    # This migration addresses a failing case where a simulation with
+    # collector nodes was not coreectly migrating the simulationType
```
dougmartin: typo: "coreectly"

src/code/data/migrations/24_fix_simulation_type.coffee
```diff
@@ -0,0 +1,18 @@
...
+    # to time-based.
+    
+    if data.nodes.some((node) -> node.data.isAccumulator)
```
dougmartin: We use lodash throughout the project (its a global, set in globals.coffee).  I'm not sure that is needed anymore but for consistency I'd replace this with:

`if _.some(data.nodes, (node) -> node.data.isAccumulator)`

test/migrations-test.coffee
```diff
@@ -150,7 +150,7 @@ describe "Migrations",  ->
     describe "v-1.23.0 changes", ->
       describe "a known failing case", ->
         # This test case is based on a field-reported bug whereby a simulation
-        # containg collector nodes was not coreectly migrating the simulationType
+        # containg collector nodes was not corectly migrating the simulationType
```
dougmartin: Close enough... :)

`containg` => `containing` collector nodes was not `corectly` => `correctly` migrating the simulationType

PR #345, opened 2018-10-17
[DL/NP]: Nodes now always have a `combineMethod` for consistent UI

knowuh:
Nodes now always have a `combineMethod` for consistent UI

* Previously it was permitted to leave this value unspecified.
However, this caused confusion because the UI was not connected to simulation behavior.
* Add migration to explicitly set the `combineMethod` for nodes without them in older models.
* A combine method is by default 'average' unless the node is
linked to a collector node (aka `target.isAccumulator` is true)
* Serialization always includes 'combineMethod` on every node.
* Node Constructor now creates `combineMethod` by default (default value `average`)
* `graphStore.getDescription` now includes `combineMethod` in stringification of targetNodes.
* Added new serialization tests.

Story:
Limiting factor is sometimes being used as default even when the setting for how to combine variables shows "average."

[#160643635]
https://www.pivotaltracker.com/story/show/160643635


test/migrations-test.coffee
```diff
@@ -159,6 +159,22 @@ describe "Migrations",  ->
         expect(brokenExample.settings.simulationType).to.equal(1) # wrong value!
         # Now migrate & test for corrected migration. 
         Migrations.update(brokenExample)
-        expect(brokenExample.version).to.equal('1.23.0')
         expect(brokenExample.settings.simulationType).to.equal(2)
 
+    describe "v-1.24.0 changes", ->
+      describe "a known failing case: missing `combineMethod` in serialization", ->
+        # We don't seem to be serializing the combinMethod
```
kswenson: typo: "combineMethod"

test/migrations-test.coffee
```diff
@@ -159,6 +159,22 @@ describe "Migrations",  ->
...
+        brokenExample = require "./serialized-test-data/v-1.22.0-missing-combine-method"
+        # Pre-test our broken json is our expected version of 1.22.0
+        expect(brokenExample.version).to.equal('1.22.0')
+        # expect to not find any `comineMethod` for our nodes.
```
kswenson: typo: "combineMethod"

PR #348, opened 2018-11-07
Added splash screen [#159109340]

dougmartin:
Changed use of git tag to a package.json version so that version shows in all builds.


src/assets/js/splash-screen.js
```diff
@@ -0,0 +1,31 @@
...
+      splashContainer.parentElement.removeChild(splashContainer);
+    };
+
+    // remove the splash screen on any click
+    splashContainer.onclick = function () {
+      clearTimeout(checkerTimeout);
+      removeSplashContainer();
+    };
+
+    var checkerTimeout;
```
knowuh: NBD, but because `checkerTimeout` is used on line 12, I was looking for `checkerTimeout` somewhere near line 4 or 5.

webpack.config.js
```diff
@@ -5,18 +5,19 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
...
       gitBranch = execSync('git symbolic-ref --short HEAD').stdout.trim(),
-      gitTags   = execSync('git describe --tags').stdout,
       gitLog = execSync('git log -1 --date=short --pretty=format:"%cd %h %ce"').stdout;
 
 const envMap = { production: "production", master: "staging" },
       environment = process.env.ENVIRONMENT || envMap[gitBranch] || "development",
       buildInfo = {
         date: buildDate,
-        tag: gitTags.split('-')[0],
+        tag: packageJSON.version,
```
knowuh: Hmn. I am wondering about this one.

git tags are a normal way we manage release versions.  With this change we would need to release using npm / `package.json`.

We don't need to address this now, bu we should revisit.

dougmartin: We can add it back but I did it so that the version displayed on the branch build.

PR #358, opened 2018-11-16
Fix image upload, fix and refactor Mixin implementation

pjanik:
1. Fixes some trivial errors related to incorrect `this` binding.
2. Previous Mixin implementation was unable to receive new `props`. I've fixed. Mixins no longer extend React.Component. Instead, they just implement their own `IMixin` interface that resembles React.Component. I think it should be a bit cleaner, as there's no need for mixin to be a "real" React Component.

That seems to fix all the errors related to image search/upload dialog.

### DIFF
--- a/src/code/mixins/components.ts
+++ b/src/code/mixins/components.ts
@@ -1,23 +1,36 @@
 import * as React from "react";
 import * as _ from "lodash";
 
-export class Mixin<P, S> extends React.Component<P, S> {
+interface IMixin<P, S> extends React.ComponentLifecycle<P, S> {
+  props: P;
+  state: S;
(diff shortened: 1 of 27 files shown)


src/code/mixins/components.ts
```diff
@@ -1,23 +1,36 @@
 import * as React from "react";
 import * as _ from "lodash";
 
-export class Mixin<P, S> extends React.Component<P, S> {
+interface IMixin<P, S> extends React.ComponentLifecycle<P, S> {
```
dougmartin: Very nice use of `extends React.ComponentLifecycle<P, S>` to ensure we are future proof from React  API changes.

PR #362, opened 2018-11-23
Simplify mini-graphs viewing, refactor migration tests, restore commented out migration tests

pjanik:
This PR implements UI changes requested in https://www.pivotaltracker.com/story/show/162022659.

Also, there's quite significant change to migration testing. IMHO it didn't work too well before.

1. Migration change data object in place. But the test data wasn't cloned. It didn't make much sense.
2. Migration tests were always running all the migrations, up to the most recent version. It was causing that old migration tests had to be deleted if the new migration was touching the same data. E.g. my recent removal of mini-graph setting was affecting initial mini-graph setting migration. I've changed tests to run migrations only to a given point, so migrations can be correctly tested and we don't have to remove any old tests.
3. Also, I've restored some commented out migration tests, as now it's possible.

### DIFF
--- a/src/code/data/migrations/26_remove_minigraphs_visibility.ts
+++ b/src/code/data/migrations/26_remove_minigraphs_visibility.ts
@@ -1,0 +1,23 @@
+/*
+ * decaffeinate suggestions:
+ * DS102: Remove unnecessary code created because of implicit returns
+ * DS207: Consider shorter variations of null checks
+ * Full docs: https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md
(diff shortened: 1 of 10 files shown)


src/code/data/migrations/26_remove_minigraphs_visibility.ts
```diff
@@ -0,0 +1,23 @@
+/*
```
dougmartin: I think you should remove this autogenerated header from the copied file.

pjanik:
Thanks, I've just removed this header so merging this PR.

PR #363, opened 2018-11-23
Reimplement splash screen, make it reusable, reorder JS loading

pjanik:
[#162022115]

I've moved HTML generation to JS file. Now we can reuse in Sage Modeler standalone site. Also, I've slightly reordered JS loading. That way splash screen shows almost immediately when the big JS files are still loading. I've tested that using Chrome network throttling.


dougmartin:
@pjanik Looks like there is a merge conflict.

pjanik:
I've fixed the conflict, so I'm merging this PR now.

PR #364, opened 2018-11-27
Let users undo node adding when Sage is embedded in CODAP

pjanik:
The main reason why we couldn't undo node add when Sage was embedded in CODAP was the fact that Sage was creating a new attribute in CODAP. This action makes CODAP document dirty and resets undo stack. When we pass `{ dirtyDocument: false }` metadata, CODAP won't mark document dirty and won't reset undo stack.

This whole undo logic is a bit questionable for me, but it works like that in other places (e.g. when we change a node). So, I'm just following an existing pattern.

There's a second part of this PR. What should we do about CODAP attribute when we undo node add and we have to remove this node?
Regular node remove action doesn't delete CODAP attribute. It makes some sense, as there might be some interesting data already. However, in the case of immediate undo action, I think it makes sense to remove the attribute, as there won't be any data. I've described that in comments. I'll double check that with Dan.

Also, this PR depends on CODAP PR: https://github.com/concord-consortium/codap/pull/262

It works without, but only partially. You'll be able to undo just a single action, as delete attribute action will reset undo stack in CODAP. When CODAP PR is merged, you'll be able to undo multiple node add actions.

### DIFF
--- a/src/code/models/codap-connect.ts
+++ b/src/code/models/codap-connect.ts
@@ -246,7 +246,10 @@ export class CodapConnect {
           const message = {
             action: "create",
             resource: `dataContext[${this.dataContextName}].collection[${this.samplesCollectionName}].attribute`,
-            values: newAttributes
+            values: newAttributes,
+            meta: {
+              dirtyDocument: false
+            }
           };
           return this.codapPhone.call(message, response => {
             if (response.success) {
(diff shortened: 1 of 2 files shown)


pjanik:
@sfentress, could you take a look at it? @dougmartin is suggesting you will be a better reviewer, as you've implemented the related code.

ddamelin:
> Regular node remove action doesn't delete CODAP attribute. It makes some sense, as there might be some interesting data already. However, in the case of immediate undo action, I think it makes sense to remove the attribute, as there won't be any data. I've described that in comments. I'll double check that with Dan.

Yes. That would be the ideal behavior.

pjanik:
Okay, thanks. So, I'm merging this PR. CODAP PR is already merged.

PR #371, opened 2019-01-09
162489262 make help (?) icon a menu

knowuh:
The (?) Icon in the menu bar now behaves as a menu with two menu items:

* About
* Help

Instead of showing the old about box, the splash screen about box (with modifications) is shown to the user when selecting (?) -> About.   

The help menu takes the user to https://concord.org/our-work/research-projects/building-models/


knowuh:
@dougmartin -- I think this is ready now.  I added a cypress test too because it was specified in the story.

PR #414, opened 2019-03-08
Fixed null is not an object on svg slider [#164496936]

dougmartin:
NOTE: this bug only occurs on Mobile Safari


dougmartin:
Fixes this Rollbar error: https://rollbar.com/concordconsortium/Sage/items/1259/

`TypeError: null is not an object (evaluating 'i.getBoundingClientRect')`

This error only occurs on mobile Safari.

You'll note in the code that I double check if the slider exists - I did that as I couldn't verify this change as I don't have a device with mobile Safari on it that will run Sage.

dougmartin:
Test url: https://sagemodeler.concord.org/?di=https://sagemodeler.concord.org/sage/branch/164496936-fix-null/sage.html

PR #421, opened 2019-03-11
Animated mini-graphs when simulation is first opened [#164435634]

dougmartin:


dougmartin:
@dstrawberrygirl Can you re-review the changes since you approved the PR?  The changes allow for the graph to animate when the experiment number is incremented.  To enable this the simulation is re-run, like it is when the simulation panel is first opened.

PR #423, opened 2019-03-12
Added "last" box checked in the parent visibility toggles [#164504977]

dougmartin:


dougmartin:
@dstrawberrygirl This adds a new parameter to the graph creation that is being merged into CODAP (more info here: https://www.pivotaltracker.com/n/projects/1055240/stories/163027761)

PR #428, opened 2019-03-15
Added Greek translation and updated help/about translations in all languages [#164610814]

dougmartin:
Also fixed missing shebang for bash in strings-pull-project.sh


dougmartin:
Default (english us):

https://sage.concord.org/branch/164610814-add-greek-trans/sage.html

New Greek translation:

https://sage.concord.org/branch/164610814-add-greek-trans/sage.html?lang=el

dougmartin:
I just looked and the "Experiment" string is not translated in any of the languages (only "Exp#" which is used in the CODAP table header).  I'll add that string to the master json and push to it POEditor and use "Πείραμα" in the Greek translation (the current greek file uses "Πείρ.#" as the translation of "Exp.#" and Google tells me that "Πείραμα" is experiment in Greek) and then pull the strings back from POEditor.

I'll also go ahead and add a title tag to all the navbar elements so if they are clipped a hover will give more info.

Finally, the `~SAGEMODELER.SHARE_DIALOG.SERVER_URL` is something used in SageModeler which shares the Sage translation files.  I noticed it missing from the master json file when I tried to merge these with SageModeler.

dougmartin:
@dstrawberrygirl I made the changes to the Greek translation PR.

dstrawberrygirl:
Looks good - I think the title hint text on clipped words is a good workaround. 👍

PR #435, opened 2019-03-19
[WIP] 164539477 animate rescale

dougmartin:
**NOTE**: ignore this for now - I just pushed this to kick off a build so that it was available during the Sage model change discussion meeting.  It needs a lot more work and needs to be rebased to master.

### DIFF
--- a/src/code/stores/graph-store.ts
+++ b/src/code/stores/graph-store.ts
@@ -96,6 +96,8 @@ export declare class GraphStoreClass extends StoreClass {
   public changeNodeWithKey(key: string, data: any): void;
   public startNodeEdit(): void;
   public endNodeEdit(): void;
+  public startNodeSliderDrag(node: Node): void;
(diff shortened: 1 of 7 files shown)


src/code/stores/graph-store.ts
```diff
@@ -666,6 +669,16 @@ export const GraphStore: GraphStoreClass = Reflux.createStore({
     return this.undoRedoManager.endCommandBatch();
   },
 
+  startNodeSliderDrag(node: Node) {
+    _.map(node.outNodes(), (outNode: Node) => outNode.unscaled = true);
```
kswenson: Since you're not using the returned array, `_.each()` seems more appropriate here and in `endNodeSliderDrag()` below.

dougmartin: Thanks.  @kswenson - I just pushed this up to make it available during a meeting looking into how the models work - its not anywhere near done.

PR #436, opened 2019-03-20
[WIP] Switch to GameView for drag/drop imports [#164761323]

dougmartin:


dougmartin:
Closing as this does not work currently in CODAP standalone mode - instead of adding the window above the Sage window it replaces it.

PR #440, opened 2019-03-22
Start  identifying graph features (model tagging)

knowuh:
This PR is mostly the inclusion of a graph analysis library [**graphLib**](https://github.com/dagrejs/graphlib).

We use this library to compute feature tags as specified in these PT stories:

* [tagging design](https://www.pivotaltracker.com/story/show/164638494) -- complete.
* [counts of model components](https://www.pivotaltracker.com/story/show/164634326) -- partial.

The analysis happens in`topology-tagger.ts` whenever the Sage model is serialized.  This `topology` field in the serialized value is incomplete at the moment, and is not read by the model.  Even so, when this work is complete, we may add a migration to reflect the new serialization schema.

We have also added a temporary hidden URL Param `?showTopology=true` and secret button for displaying the topology in order to demonstrate these features to PIs while the feature is under development. These features will be removed once we have a better way to report on the data.

More work remains to be done in extracting and tagging specific features as per several PT Stories:
* [specific loop types](https://www.pivotaltracker.com/story/show/164634579)
* [linear models](https://www.pivotaltracker.com/story/show/164634474)  
* &etc.

@dougmartin could you please give us a review?

### DIFF
--- a/src/code/utils/topology-tagger.ts
+++ b/src/code/utils/topology-tagger.ts
@@ -1,31 +1,34 @@
+const _ = require("lodash");
+import { Graph as AnalysisGraph, alg as GraphLibAlg} from "graphlib";
+import { Link } from "../models/link";
+import { Node } from "../models/node";
 
-import { Graph, alg} from "graphlib";
+type sageNodeKey = string;
 
-export interface ILinkStruct {
-  sourceNode: string;
-  targetNode: string;
+export interface ISageLink {
+  sourceNode: sageNodeKey;
+  targetNode: sageNodeKey;
   title: string;
 }
 
-export interface InodeDataStruct {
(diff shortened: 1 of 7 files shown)


src/code/utils/topology-tagger.ts
```diff
@@ -0,0 +1,54 @@
...
+export function countNodes(g: AnalysisGraph) {
+  return g.nodeCount();
+}
+
+export function getCycles(g: AnalysisGraph) {
+  return GraphLibAlg.findCycles(g);
+}
+
+
+export function getToplogy(sageModelGraph: ISageGraph) {
```
dougmartin: Missing an "o": `getToplogy` => `getTopology`

PR #441, opened 2019-03-22
Add CODAP guide UI [#161954869]

dougmartin:
Everything except for the CODAP API call to switch guide pages is working.

### DIFF
--- a/src/code/models/codap-connect.ts
+++ b/src/code/models/codap-connect.ts
@@ -18,6 +18,7 @@ import { PaletteStore } from "../stores/palette-store";
 import { TimeUnits } from "../utils/time-units";
 import { escapeRegExp } from "../utils/escape-reg-ex";
 import { GraphStore, GraphStoreClass } from "../stores/graph-store";
+import { AppSettingsActions } from "../stores/app-settings-store";
 
 // log -- see loglevel in package.json
 
@@ -67,6 +68,8 @@ export class CodapConnect {
   private tableCreated: boolean;
   private _attrsToSync: any; // checked: any ok
   private _attrsAreLoaded: boolean;
+  private guidePollerInterval: number | null = null;
+  private guideComponentId: number | null = null;
 
   constructor(context) {
     this.codapRequestHandler = this.codapRequestHandler.bind(this);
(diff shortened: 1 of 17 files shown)


dougmartin:
@knowuh / @DALove1025 - this is ready for review but not for merging.  The CODAP API call to switch guide pages is not yet in master or production.

src/code/models/codap-connect.ts
```diff
@@ -780,6 +799,61 @@ export class CodapConnect {
...
+        }
+      });
+    } else {
+      getGuideComponent(this.guideComponentId);
+    }
+  }
+
+  private pollForGuideItemUpdates() {
+    if (!this.guidePollerInterval) {
+      this.guidePollerInterval = window.setInterval(() => this.getGuideItems(), 1000);
```
knowuh: Not knowing how this API is meant to be used, I have a question:  Could the guideItems change?  Do we have to poll forever, or should we eventually be clearing the interval?

src/code/stores/app-settings-store.ts
```diff
@@ -148,7 +171,8 @@ export const AppSettingsStore: AppSettingsStoreClass = Reflux.createStore({
     return {
       complexity: this.settings.complexity,
       simulationType: this.settings.simulationType,
-      relationshipSymbols: this.settings.relationshipSymbols
+      relationshipSymbols: this.settings.relationshipSymbols,
```
knowuh: I should go read what these `relationshipSymbols` are all about.

src/code/models/codap-connect.ts
dougmartin: We have to poll forever once the guide configuration gui is opened - CODAP doesn't give us any callbacks to find out when the gui is closed.  I setup the poller this way because it will be a rare occurance to configure the guides so the poller is only enabled when the gui is opened.

kswenson: Did you discuss this with @jsandoe ? A notification on close seems like a reasonable request for addition to the plugin API.

dougmartin:
Don't merge now as CODAP causes the sage navbar to scroll out of view when the guide window is opened.

PR #447, opened 2019-03-25
Upgraded to React 16 [#161154348]

dougmartin:
Test deployment: https://sage.concord.org/branch/161154348-update-to-react-16/sage.html


test/value-slider-view-test.ts
```diff
@@ -38,6 +43,14 @@ describe("The Value Slider", () => {
     });
   };
 
+  // fake the event listeners for enzyme
```
knowuh: Just curious:  What is the story with this?

dougmartin: Enzyme doesn't support `window.addEventListener()` and its only used to check for resizing which the test doesn't exercise.  The docs say to just stub it out.

PR #453, opened 2019-03-27
Add nudge via keyboard of current slider [#164895490]

dougmartin:
Currently a WIP branch as the final default nudge factor needs to be decided, otherwise the code is finalized.


dougmartin:
Demo branch: https://sage.concord.org/branch/164833716-keyboard-slider/sage.html?nudgeFactor=1#shared=94180

PR #454, opened 2019-03-27
Make the white area of a selected node draggable

dstrawberrygirl:
Addresses [#162094889] to make the white area behind a selected node draggable, not just the node image itself. Implemented a solution involving a temporary extra div that is the same size as the node while the node is selected, raised the link button up a notch on z-index to ensure that functionality was not altered while the selected mode was in effect.


dougmartin:
**DON'T MERGE** - there is a bug with this code noted in PT here: https://www.pivotaltracker.com/story/show/162094889/comments/201104569

PR #456, opened 2019-03-28
Implemented branch detection & unit tests

DALove1025:
PR for [SageModeler generates tag for branched model](https://www.pivotaltracker.com/story/show/164634570).

This is for the simple cases. Will have to be revisited depending on how we decide to handle transfer links.


test/topology-tagger-test.ts
```diff
@@ -200,9 +200,60 @@ describe("A TopologyTagger", () => {
...
+          {key: "c"}
+        ];
+        const links: ISageLink[] = [
+          {title: "L1", sourceNode: "a", targetNode: "b"},
+          {title: "L2", sourceNode: "a", targetNode: "c"}
+        ];
+        const graph: ISageGraph = { nodes, links };
+        chai.expect(getTopology(graph).isBranched).to.eql(true);
+      });
+      it("detects a join from two nodes to one", () => {
```
knowuh: It's probably worth clarifying with Dan if a join is qualifies as a branch.

src/code/utils/topology-tagger.ts
```diff
@@ -78,6 +78,13 @@ function isLinearModel(g) {
     .reduce ( (a, b) => (a + b), 0) === g.nodeCount();
 }
 
+function anyBranches(g) {
+  const brancedOrJoinedNodeCount = g.nodes()
+    .map( (node) => ((g.inEdges(node).length > 1 || g.outEdges(node).length > 1 ) ? 1 : 0))
```
knowuh: There is an opportunity to bail early if we use `Array.find()` instead of `Array.map()`

But it's probably an unimportant optimization at this point.

PR #457, opened 2019-03-29
Added buttons to nudge value slider [#164895490]

dougmartin:
PT: https://www.pivotaltracker.com/story/show/164895490

Demo: https://sage.concord.org/branch/164895490-button-slider/sage.html?wiggleLookback=10&wiggleThreshold=0.2

### DIFF
--- a/src/code/views/value-slider-view.tsx
+++ b/src/code/views/value-slider-view.tsx
@@ -322,6 +322,7 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
           max = Math.max(max, lookbackValue);
         }
         if (max - min <= maxChange) {
+          debugger;
           this.setState({showButtons: true});
         }
       }
@@ -605,6 +606,8 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
   }
 
   private handleNudge = (delta: number) => {
+    const {min, max} = this.props;
+    const maxNudge = (max - min) * 0.5;
     const startOfNudge = Date.now();
     const clearMouseDownInterval = () => {
       if (this.nudgeInterval) {
@@ -621,8 +624,8 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
       const now = Date.now();
       // create an exponent for the modifer of number of seconds after 1 second from the start of the nudge
       const exp = Math.max(0, (now - startOfNudge - 1000) / 1000);
-      // create modifier starting after 1 second (Math.pow(2, 0) is 1 which is why 1 is subtracted)
-      const mod = Math.pow(2, exp) - 1;
+      // create modifier starting after 1 second (Math.pow(10, 0) is 1 which is why 1 is subtracted)
+      const mod = Math.min(maxNudge, Math.pow(10, exp) - 1);
       nudge(delta + (delta * mod));
     };
     const mouseUp = (e: MouseEvent) => {
--- a/src/code/views/value-slider-view.tsx
+++ b/src/code/views/value-slider-view.tsx
@@ -322,6 +322,15 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
           maxAbsDelta = Math.max(maxAbsDelta, Math.abs(wiggleList[i] - oldestValue));
         }
         if (maxAbsDelta <= maxChange) {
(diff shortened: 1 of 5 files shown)


src/code/views/value-slider-view.tsx
```diff
@@ -1,8 +1,18 @@
 import * as React from "react";
 const log = require("loglevel");
 
+import { urlParams } from "../utils/url-params";
 import { tr } from "../utils/translate";
 
+const urlValueWithDefault = (name: string, defaultValue: number, urlValue?: string) => {
```
knowuh: Seems like this function `urlValueWithDefault` should exist in `urlParams` util package?

src/code/views/value-slider-view.tsx
```diff
@@ -217,12 +236,33 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
     return <div className="number" style={style}>{this.props.value.toFixed(this.props.displayPrecision)}</div>;
   }
 
+  private isButtonEvent(e) {
+    const {target} = e;
+    const {parentNode} = target;
+    return (target.getAttribute("class") === "slider-button") || (parentNode && (parentNode.getAttribute("class") === "slider-button"));
+  }
+
   private handleNoop = (e) => {
```
knowuh: I know that this commit and PR doesn't modify this method `handleNoop` -- but when I was reading your code it occurred to me that this is a bad name for this method.

The method name should be like `stopPropogation` or something.

src/code/views/value-slider-view.tsx
```diff
@@ -485,4 +559,81 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
       }
     }
   }
+
+  private toggleButtons() {
+    this.setState({showButtons: !this.state.showButtons});
+  }
+
+  private renderButtons() {
```
knowuh: These wiggle-toggle-buttons are complex enough to warrant their own component I think.

src/code/views/value-slider-view.tsx
```diff
@@ -240,21 +283,51 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
...
     this.handleDrag(e);
     return this.handleStart(e);
   }
 
+  // shows the buttons if the user moves the slider around a small range
+  private checkForSliderWiggle(value: number) {
+    if (!this.state.showButtons) {
+      const {wiggleList} = this;
+      const {length} = wiggleList;
+      wiggleList.push(value);
```
knowuh: I am not seeing where this list gets trimmed.

src/code/views/value-slider-view.tsx
```diff
@@ -240,21 +283,51 @@ export class SVGSliderView extends React.Component<SVGSliderViewProps, SVGSlider
...
+      if (length >= SliderWiggleLookback) {
```
knowuh: I might need a description of what we are doing in this loop. I got lost.

1. We are looking at the last `n` changes in absolute model `value` history (`wiggleList`).
2. If the maximum span (between the lowest and highest value) in our list is <= `maxChange` then we show the buttons.

I guess that makes sense.  `maxChange` is a threshold value, if we go over it, we should also `setState({showButtons: false})` I would think?  I am not sure  what to call `maxChange` but the name is a little confusing to me.

src/code/views/value-slider-view.tsx
dougmartin: I shouldn't have asked for a review for this PR yet - this function was in to allow the building models folks to tweak the final internal settings without me rebuilding it - this will all be removed.

src/code/views/value-slider-view.tsx
dougmartin: I agree, or `stopEvent()`

src/code/views/value-slider-view.tsx
dougmartin: It gets reset on the click event handler which is the last event called.

src/code/views/value-slider-view.tsx
dougmartin: This is meant to show the buttons when the user is "wiggling" around a value but not hide them if they are already showing.  After playing with it with Dan on Friday its not exactly what he wants.  I'm going to change it today to store deltas and see if that gets closer to the behavior he wants.

PR #528, opened 2019-09-19
Add wait for graphstore before connecting to CODAP

dougmartin:
The latest CODAP is faster at returning messages after connecting and we've found a race condition between the undoRedoManager becoming available and the returned messages causing methods to be called on the undoRedoManager.  This PR waits for the graph store to become available (ie the undoRedoManager is inited) before talking to CODAP.

### DIFF
--- a/src/code/models/codap-connect.ts
+++ b/src/code/models/codap-connect.ts
@@ -91,75 +91,78 @@ export class CodapConnect {
     this.codapPhone = new IframePhoneRpcEndpoint( this.codapRequestHandler,
       "data-interactive", window.parent );
 
-    // load any previous data; also check if CODAP's undo is available,
-    // or if we are in standalone mode.
-    this.codapPhone.call([
-      {
-        action: "update",
-        resource: "interactiveFrame",
-        values: {
-          title: tr("~CODAP.INTERACTIVE_FRAME.TITLE"),
-          preventBringToFront: true,
-          cannotClose: true
+    // wait for the graphstore to be ready to receive the CODAP data
+    this.graphStore.waitUntilReady(() => {
+      // load any previous data; also check if CODAP's undo is available,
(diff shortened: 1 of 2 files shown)


src/code/models/codap-connect.ts
```diff
@@ -91,75 +91,78 @@ export class CodapConnect {
...
+          context = ret[2];
+
+          this.graphStore.setUsingCODAP(true);
+
+          if (frame != null ? frame.values.externalUndoAvailable : undefined) {
+            CodapActions.hideUndoRedo();
+          } else if (frame != null ? frame.values.standaloneUndoModeAvailable : undefined) {
+            this.standaloneMode = true;
+            this.graphStore.setCodapStandaloneMode(true);
+          }
```
knowuh: The display of this diff on github is pretty bad :)

src/code/models/codap-connect.ts
```diff
@@ -91,75 +91,78 @@ export class CodapConnect {
...
+        resource: `dataContext[${this.dataContextName}]`
       }
-      this.updateExperimentColumn();
-      return this._getExperimentNumber();
+      , ret => {
+        // ret==null is indication of timeout, not an indication that the data set
+        // doesn't exist.
+        if (!ret || ret.success) {
+          let attrs;
+          if (attrs = (__guard__(__guard__(__guard__(ret != null ? ret.values : undefined, x2 => x2.collections), x1 => x1[1]), x => x.attrs) != null)) {
```
knowuh: I realize this isn't new code but this is ... something. 

Assignment in a conditional evaluation, assigned to a tertiary with function calls such as `(__guard__(__guard__(__guard__(`   

Not a fan.

dougmartin: This is all cruft over from the CoffeScript -> TypeScript conversion.

PR #550, opened 2019-11-19
Fixed can't edit variable name when node is in selected state [#169842945]

dougmartin:


dougmartin:
@dstrawberrygirl there is a draggble div "under" the node-title-box but it is absolutely positioned so it changes the z-index of the divs after it.  This change makes the node-title-box also absolutely positioned and then forces it above the draggable div in z order.

PR #552, opened 2019-11-20
Update translated "Help" menu

dougmartin:


dougmartin:
@dstrawberrygirl Dan added translations for "Help" in Norwegian and Greek.

PR #555, opened 2019-11-22
Change model settings text and layout [#169606086]

dougmartin:


dougmartin:
Previous layout:

image

New layout:

image

dougmartin:
Added WIP label to prevent merge.  Still need to address comment by Dan.

dougmartin:
@dstrawberrygirl could you review this added commit?

https://github.com/concord-consortium/building-models/pull/555/commits/fe0e34065d5ae71c03417b37f9f11a25690a3e65

NOTE: ignore the `TODO: SWITCH ONLY IF ALL LINKS ARE NOT DEFINED` comment that was removed.  I added that in the previous commit (which wasn't pushed) to mark where I needed to fix this.  I should have rebased and removed that commit.

PR #556, opened 2019-11-25
Fix csv drops into CODAP [#168381157]

dougmartin:
Disables the componentType when importing csv data into CODAP so that csv import path can be followed.  Also adds a name value so CODAP can use it instead of the contents of the data-uri for the table name.


dougmartin:
@dstrawberrygirl this works hand in hand with this CODAP PR: https://github.com/concord-consortium/codap/pull/299 but does not require it to be deployed to work.

PR #557, opened 2019-11-26
Change new table to use CODAP API [#169978832]

dougmartin:
Previously it imported a csv constructed of a single column with no data but this triggers unwanted behavior in the updated CODAP csv importer.

### DIFF
--- a/src/code/models/codap-connect.ts
+++ b/src/code/models/codap-connect.ts
@@ -838,7 +838,7 @@ export class CodapConnect {
   }
 
   public addNewDataContextAndTable() {
-    const newDataContextName = `NewTable${Math.floor(10000000000 * Math.random())}`;
+    const newDataContextName = `NewTable${Date.now()}`;
     this.codapPhone.call({
       action: "create",
       resource: "dataContext",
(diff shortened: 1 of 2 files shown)


src/code/models/codap-connect.ts
```diff
@@ -837,6 +837,35 @@ export class CodapConnect {
     }
   }
 
+  public addNewDataContextAndTable() {
+    const newDataContextName = `NewTable${Math.floor(10000000000 * Math.random())}`;
```
dstrawberrygirl: If you're going for a unique-ish identifier then my personal preference is something like current date/time as ticks.

src/code/models/codap-connect.ts
```diff
@@ -837,6 +837,35 @@ export class CodapConnect {
...
+        }]
+      }
+    }, (result) => {
+      this.codapPhone.call({
+        action: "create",
+        resource: "component",
+        values: {
+          type: "caseTable",
+          title: tr("~DOCUMENT.CODAP_ACTIONS.TABLES.NEW"),
+          dataContext: newDataContextName,
```
dstrawberrygirl: There's an assumption here that the call to create the dataContext was successful - could this fail? We never use any of the information returned in the result.

dougmartin:
@dstrawberrygirl I updated the PR based on your comments.

PR #568, opened 2019-12-05
Added script to compare language files [#170116509]

dougmartin:
This will allow us to see what is not translated and if there are missing keys.


dougmartin:
No review necessary - this is a utlity to use while the internationalization stories are worked on.

PR #571, opened 2019-12-09
Added translation strings for image metadata

dougmartin:


dougmartin:
@dstrawberrygirl removing review request on this one - there is one more section of translations I need to do.

PR #574, opened 2019-12-13
Remove Polish and Estonian as CODAP doesn't support them

dougmartin:


dougmartin:
@dstrawberrygirl I reassigned these to Piotr since he is now working on Sage.

PR #578, opened 2019-12-20
Disabled topology-tagger [#170410042]

dougmartin:
Temporary solution to remove breaking bug.

### DIFF
--- a/src/code/stores/graph-store.ts
+++ b/src/code/stores/graph-store.ts
@@ -965,7 +965,8 @@ export const GraphStore: GraphStoreClass = Reflux.createStore({
     })();
     const settings = AppSettingsStore.serialize();
     settings.simulation = SimulationStore.serialize();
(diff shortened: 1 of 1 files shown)


src/code/stores/graph-store.ts
```diff
@@ -965,15 +965,16 @@ export const GraphStore: GraphStoreClass = Reflux.createStore({
...
-    const topology = getTopology({nodes: nodeExports, links: linkExports});
+    // TODO: fix and re-enable https://www.pivotaltracker.com/n/projects/1263626/stories/170410042
+    // const topology = getTopology({nodes: nodeExports, links: linkExports});
     const data = {
       version: latestVersion(),
       filename: this.filename,
       palette,
       nodes: nodeExports,
       links: linkExports,
       settings,
```
knowuh: Trailing comma?

dougmartin: The Typescript compiler rewrites all of this so it doesn't matter.

PR #580, opened 2020-01-03
Add Simulation V2 which provides RK4 integration method

pjanik:
This PR adds a new module called Simulation V2. It's partially based on previous Simulation class (which is still there but renamed to SimulationV1), although there are many changes:

1. No hacks with 10 pre-steps and 30 steps that average values (unless it's necessary, more about that below). Cleaner integration loop.
1. Static nodes are evaluated recursively and immediately. We used to treat them as a part of the integration loop, and each step included 10 "pre-steps" to push values down the chain. This didn't seem correct and was limiting chains of static relationships to 10 nodes, or there was some delay.
1. Loops that include only static nodes are now detected and evaluated only if necessary. A similar method is used -> 30 steps + averaging. But it's more explicit and not run all the time, even if it's not necessary.
1. The Euler method is more explicitly implemented.
1. The RK4 method is also added.
1. Time step is now called timeStep and there are some bug fixes related to its handling (e.g. transfer nodes didn't use it before).
1. Changing the time step to a value smaller than 1 doesn't require the user to increase the number of steps in the simulation settings. It will happen automatically. But we'll still output data only at whole number time values. So, changing the time step will only increase accuracy and decrease performance. It won't have any other visible effects for the user.
1. I reused SimulationV1 tests. Although I had to change some details, plus I've added tests of the RK4 method and different time steps.

### DIFF
--- a/src/code/models/node.ts
+++ b/src/code/models/node.ts
@@ -43,7 +43,7 @@ export class Node extends GraphPrimitive {
 
   public type: string = "Node";
   public title: string;
-  public currentValue: number | null;
+  public currentValue: number;
 
   public readonly combineMethod: any; // TODO: get concrete type
   public readonly valueDefinedSemiQuantitatively: any; // TODO: get concrete type
@@ -207,11 +207,11 @@ export class Node extends GraphPrimitive {
     }
   }
 
-  public outLinks(relationType = null) {
+  public outLinks(relationType: string | null = null) {
     return _.filter(this.links, link => (link.sourceNode === this) && ((relationType === null) || (relationType === link.relation.type)));
   }
 
-  public inLinks(relationType = null) {
+  public inLinks(relationType: string | null = null) {
     return _.filter(this.links, link => (link.targetNode === this) && ((relationType === null) || (relationType === link.relation.type)));
   }
 
--- a/src/code/models/node.ts
+++ b/src/code/models/node.ts
@@ -43,7 +43,7 @@ export class Node extends GraphPrimitive {
 
   public type: string = "Node";
   public title: string;
-  public currentValue: number;
+  public currentValue: number | null;
 
   public readonly combineMethod: any; // TODO: get concrete type
   public readonly valueDefinedSemiQuantitatively: any; // TODO: get concrete type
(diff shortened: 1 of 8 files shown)


dougmartin:
I think @sfentress would be a better person to review this as he built the first version of the simulation engine.

pjanik:
Thanks!

> In your replacement #4 scenario test, A <-> [B], where B is set to 50, I expect in one step A to be set to 50 and fed back to B, making 100. 

This test case is generally confusing and I wasn't sure if it's worth keeping it. Note that this feedback loop generally causes two nodes to go to infinity. If we had absolutely perfect integration method with infinitely small time step, values at any time different from 0 should equal to infinity. But we don't, so we get different values for different time steps and integration methods.

Let's analyze Euler, as it's simpler. Initially, both nodes are set to 50 (as A = B).

1. When the time step is equal to 1, we add 1 * 50 to B during the first step. So it's 100. A is also updated to 100. So, values at t=1 are 100 and 100.
2. When the time step is equal to 0.5, we add 0.5 * 50 to B, so it's 75. A is updated to 75. So, values at t=0.5 are 75 and 75. We need a second step to get to t=1. So, we add 0.5 * 75 to B, so it's 112.5. A is also updated to 112.5. So, values at t=1 are 112.5 and 112.5. Bigger than 100 and 100 that we got for dt=1. 
3. If time step was 0.25, ..., I guess you get the idea. Values would be bigger again.

RK4 method performs a few calculations for different time steps and averages results (well, in a big simplification), so we get a bit closer to infinity too. Which is nice in some sense. ;) 

I left this test, as it's the only synthetic test that shows some differences between Euler and RK4 precision. Others are so simple that Euler's poorer precision is not visible (but at least we test that RK4 returns the same results).

This model shows Euler vs RK4 difference very clearly:
1. Euler: https://sagemodeler.concord.org/app/?di=170288915-sim-v2&sage:simulation=v1#shared=135562
2. RK4: https://sagemodeler.concord.org/app/?di=170288915-sim-v2&sage:simulation=v2&sage:integration=rk4#shared=135562

Does it make it a bit more clear?

SamFen:
Right, yes that's what I was getting at. Thanks for the explanation.

PR #588, opened 2020-01-21
Fixed crashing bug when transfer link reasoning was updated [#170805723]

dougmartin:


dougmartin:
@pjanik yes - if no relation was passed into the function in the set of link changes it removed the transfer node from the link.  The fix was to first check if the relation changed and then add/remove the transfer node based on the relation.

PR #610, opened 2020-04-01
Image collections

sfentress:
This makes it possible to author collections of images that appear in the New Image dialog. These are defined with JSON files and passed in via a url parameter.

This can be demoed here: http://sage.concord.org/branch/image-collections/sage.html?imageCollections=image-collections/cbio/collections/population-cell.json

This is the dialog that appears when you press the New Image button:

Screen Shot 2020-03-31 at 9 33 01 PM

The actual JSON and image files are for demo purposes only, but it seems useful to keep them in the repo for testing.

This also fixed a couple small image dialog issues: the selected tab being cleared whenever you clicked on an image (not typically noticeable previously because it only happened on the Upload or Link tabs), and clearing the selected image when a new tab is selected.


src/code/stores/palette-store.ts
```diff
@@ -85,6 +89,57 @@ export const PaletteStore: PaletteStoreClass = Reflux.createStore({
...
+        log.info(`loading collection ${url}`);
+        return $.ajax({
+          url,
+          dataType: "json",
+          success: data => {
+            if (data.collections) {   // nested collection
+              // first replace our current url in collectionsOrder with our new list
+              collectionsOrder.splice(collectionsOrder.indexOf(url), 1, ...data.collections);
+              // then recursively load nested collections
+              data.collections.forEach(loadCollection);
```
dougmartin, line 128: We have been using `lodash` for for loops on arrays based on older browsers not supporting that but I think enough time has passed that using the built in functions is fine now.

src/code/views/tabbed-panel-view.tsx
```diff
@@ -37,46 +41,37 @@ class TabbedPanelTabView extends React.Component<TabbedPanelTabViewProps, Tabbed
...
-export class TabbedPanelView extends React.Component<TabbedPanelViewProps, TabbedPanelViewState> {
+export class TabbedPanelView extends React.Component<TabbedPanelViewProps> {
 
   public static displayName = "TabbedPanelView";
 
   public static Tab(settings?: TabInfoSettings) {
     return new TabInfo(settings);
   }
 
-  constructor(props: TabbedPanelViewProps) {
```
dougmartin, line 69: Not sure why we where tracking our own state version of the prop so this change seems like a good idea.  I'll be curious if this exposes some other issue on another use of the tabbed panel where we needed to track state for some reason.

src/code/stores/palette-store.ts
sfentress: Hmmm, `forEach` has been [supported since IE 9](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Browser_compatibility) (and Chrome 1, Firefox 1.5, etc.). Was that ever really a concern for us?

src/code/views/tabbed-panel-view.tsx
sfentress: The only other use of this view was the relationship panel, which used the props-version of the selected tab. Yeah, it seemed like a good opportunity to fix an anti-pattern.

PR #612, opened 2020-04-14
Add all to palette button for image collections

sfentress:
Can be tested here: http://sage.concord.org/branch/172180682-add-all-to-palette/sage.html?imageCollections=https://models-resources.concord.org/connected-bio-resources/population-cell-protein-synthesis-heredity.json

### DIFF
--- a/src/code/utils/undo-redo.ts
+++ b/src/code/utils/undo-redo.ts
@@ -72,6 +72,7 @@ export class UndoRedoManager {
       if (this.currentBatch.commands.length > 0) {
         this.commands.push(this.currentBatch);
         this.stackPosition++;
+        this._changed();
       }
       return this.currentBatch = null;
     }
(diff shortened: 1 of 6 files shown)


src/code/utils/undo-redo.ts
```diff
@@ -72,6 +72,7 @@ export class UndoRedoManager {
       if (this.currentBatch.commands.length > 0) {
         this.commands.push(this.currentBatch);
         this.stackPosition++;
+        this._changed();
```
dougmartin, line 75: I'm surprised this missing didn't cause issues before this.

sfentress: The only thing it does is trigger the UI update. And virtually anything the user does after that will trigger the same update. And, 99% of the time, the UI isn't actually expected to change, because it's just whether the Undo/Redo arrows show that there is something to undo, and usually there is.

This was only noticeable because I was testing  this feature by adding all images as my very first action, and noticing that the Undo button still appeared to be disabled (even though it still actually worked).

PR #613, opened 2020-04-17
Add optional fullscreen button (?fullscreenButton=true)

sfentress:
When Sage is embedded using the Document Store, a fullscreen button is provided for it, but otherwise it doesn't have one.

This adds a url parameter "fullscreenButton" that can provide a similar-looking fullscreen button.

https://www.pivotaltracker.com/story/show/172377261

### DIFF
--- a/src/assets/img/fullscreen-exit.svg
+++ b/src/assets/img/fullscreen-exit.svg
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 width="20px" height="20px" viewBox="0 0 20 20" enable-background="new 0 0 20 20" xml:space="preserve">
+  <g id="Layer_1">
+    <g>
+      <polyline fill="none" stroke="#000" stroke-width="2" stroke-miterlimit="10" points="18,14 13,14
+        13,19 		"/>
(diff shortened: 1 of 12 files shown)


src/code/views/fullscreen-button.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export class FullScreenButton extends React.Component<IProps, IState> {
+
+  public state = {
+    isFullscreen: false
+  };
+
+  public componentDidMount() {
+    if (screenfull.isEnabled) {
+      screenfull.on("change", () => {
```
dougmartin, line 22: Looks like you need to add a `componentWillUnmount` method and call `screenfull.off("change", ...)` and move the callback out into the class.

https://github.com/sindresorhus/screenfull.js/#detect-fullscreen-change

src/assets/img/fullscreen-exit.svg
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dougmartin, line 1: Are these the same buttons as in Lara?  Kiley wants to ensure we use the same buttons in the UI.

src/code/views/fullscreen-button.tsx
sfentress: :+1:

src/assets/img/fullscreen-exit.svg
sfentress: Yes, they are.

sfentress:
@dougmartin New changes added. The handler was removed on unmount, and a scaling parameter was added which works identically to the way it does in Document Store. It can be tested at http://sage.concord.org/branch/add-fullscreen-button/sage.html?fullscreenButton=true&scaling=true

PR #614, opened 2020-04-22
Fix scaling of app

sfentress:
When using stand-alone Sage scaling, we have to handle all the mouse transformations ourselves. This wasn't an issue in the DocumentStore, because that scaled the entire iframe, which allows the mouse coordinates of the internal app to ignore scaling.

This is mostly straight-forward, dividing coordinates as needed by the current scale. The only annoying thing (which caused hours of face-desking) is how jsPlumb handles zooming. jsPlumb does have a `setZoom` function, which would in theory work, but each node has a drag-drop handler which doesn't hear the `zoom` event, because we call `reset` after the handler's listener is created. This hacks jsPlumb to use a single top-level zoom scale in this case.

No behavior should be different if scaling is 1 (which is always the case is `?scaling=true` is not set).

As far as I can tell, this covers all the cases, but I will not merge this in until it passes QA.

### DIFF
--- a/src/code/utils/js-plumb-diagram-toolkit.ts
+++ b/src/code/utils/js-plumb-diagram-toolkit.ts
@@ -12,7 +12,7 @@ const log = require("loglevel");
 import { LinkColors } from "../utils/link-colors";
 import * as $ from "jquery";
 
-import { getViewScale } from "../utils/scale-app";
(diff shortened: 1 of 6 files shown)


ddamelin:
QA should include testing that the full SageModeler app embedded in LARA using scaling is not affected.

PR #622, opened 2020-07-21
Added log events [#172062598]

dougmartin:


src/code/views/node-view.tsx
```diff
@@ -265,6 +266,7 @@ export class NodeView extends React.Component<NodeViewProps, NodeViewState> {
   private lastClickLinkTime;
   private initialTitle: string;
   private node: HTMLDivElement | null;
+  private logValueLogTimeout;
```
knowuh, line 269: I don't see this value being used anywhere …

dougmartin: Good catch.  I moved the code that used that into the graph view.  I've added a commit to remove the line.

PR #623, opened 2020-07-23
Fixed building-models side of dirtying document on load [#170849917]

dougmartin:
Adds a flag to skip undo/redo when importing nodes so that CODAP isn't notified of the load.  Without this flag CODAP thinks the document is getting dirtied at load time.

### DIFF
--- a/src/code/stores/graph-store.ts
+++ b/src/code/stores/graph-store.ts
@@ -53,6 +53,10 @@ interface LogOptions {
   logEvent?: boolean;
 }
 
+interface UndoRedoOptions {
+  skipUndoRedo?: boolean;
+}
+
 export declare class GraphStoreClass extends StoreClass {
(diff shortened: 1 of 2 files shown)


src/code/stores/graph-store.ts
```diff
@@ -53,6 +53,10 @@ interface LogOptions {
   logEvent?: boolean;
 }
 
+interface UndoRedoOptions {
+  skipUndoRedo?: boolean;
+}
+
```
kswenson, line 59:
```suggestion
type RecordOptions = LogOptions & UndoRedoOptions;
```
Just to DRY up the code below.

dougmartin: I kind of like it having the two types in the method signatures - it feels a little more self-documenting to me.  I'm going to leave it in for now.

PR #639, opened 2020-12-11
Added preventDataContextReorg to CODAP plugin params [#170897787]

dougmartin:
Changes preventDataContextReorg property of SageModeler plugin to be "false" so that tables can be reorganized, including those new tables that get created through the "new table" button.


dstrawberrygirl (approved):
Looks good!

PR #644, opened 2020-12-20
Added naming of flow [#171530932]

dougmartin:
- When a flow variable is created it still gets the same default name and this name is hidden.
- If the user hovers over or selects a flow variable, the name becomes visible _and_ editable.
- If the user edits the variable name, then this name is now shown even when the variable is not selected or hovered over. In other words, it behaves just like any other variable.
- If the user delete the customized name of the flow variable, it reverts to its default name and hidden behavior.


pjanik (approved):
Looks good to me 👍

PR #645, opened 2020-12-23
Fixed can't draw vary line in touchscreen [#159155537]

dougmartin:
Also sped up event handling by using passive flag on mouse events.


pjanik (approved):
Looks good 👍

PR #648, opened 2021-01-15
Guard against LARA passing empty string

sfentress:
If a model has no authoredState, and the model is being used in runtime, LARA will pass in an initial state as 

```
{
    ...
    authoredState: null
}
```

which is fine.

However, for some reason in report-mode, LARA passes the initial state as

```
{
    ...
    authoredState: ""
}
```

which causes `JSON.parse` to throw an error. This adds a simple check on length which should fix that, and also protects it further with try/catch.

(This should also be fixed on LARA's end, but that's out of scope right now.)


dougmartin (approved):
Looks good :+1:

PR #652, opened 2021-01-29
Version 2.3.1

sfentress:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "building-models",
   "description": "System dynamics modeling tool.",
   "author": "The Concord Consortium",
-  "version": "2.3.0",
+  "version": "2.3.1",
```
kswenson, line 5: Should probably update `package-lock.json` as well.

sfentress: :+1:

PR #663, opened 2021-04-30
Updated plural week translation for Norwegian

dougmartin:


dougmartin:
Going to merge this without review as it is a one line translation change.

PR #675, opened 2021-06-02
Collapsed slider dragging actions into one undoable step [#171509259]

dougmartin:
This adds both the wrapping of a command batch around the slider start/end drag and a new setter method in the undoRedo manager to enable/disable further command batches.  This needed as each slider value change caused a seperate command batch to be created.

This also fixes a typo in the internal end command batch method.

### DIFF
--- a/src/code/utils/undo-redo.ts
+++ b/src/code/utils/undo-redo.ts
@@ -36,6 +36,7 @@ export class UndoRedoManager {
   private changeListeners: UndoRedoListener[];
   private currentBatch: UndoRedoCommandBatch | null;
   private debug: boolean;
+  private commandBatchingEnabled: boolean;
 
   constructor(options) {
     this.endCommandBatch = Reflux.createAction();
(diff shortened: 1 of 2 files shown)


src/code/utils/undo-redo.ts
```diff
@@ -49,9 +50,10 @@ export class UndoRedoManager {
     this.savePosition = -1;
     this.changeListeners = [];
     this.currentBatch = null;
+    this.commandBatchingEnabled = true;
 
     // listen to all our actions
-    this.endCommandBatch.listen(this._endComandBatch, this);
+    this.endCommandBatch.listen(this._endCommandBatch, this);
```
knowuh, line 56: Errors like this typo make me wonder why/how software works at all.  Nice catch, and how did things ever work before?

knowuh (approved):
Thanks, looks good.

src/code/utils/undo-redo.ts
dougmartin: The `_endComandBatch` was a private method not called outside the class wrapped by a public `endCommandBatch` method so it did all work before, but I thought since I saw it it would be good to fix.

PR #678, opened 2021-06-03
Fixed saving of dropped image urls or local image files [#170858656]

dougmartin:
This also adds the imge to image pallete as requested in the story.

### DIFF
--- a/src/code/models/node.ts
+++ b/src/code/models/node.ts
@@ -361,7 +361,8 @@ export class Node extends GraphPrimitive {
         allowNegativeValues: this.allowNegativeValues,
         valueDefinedSemiQuantitatively: this.valueDefinedSemiQuantitatively,
         frames: _.clone(this.frames),
-        combineMethod: this.combineMethod
+        combineMethod: this.combineMethod,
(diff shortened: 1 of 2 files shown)


src/code/models/node.ts
```diff
@@ -361,7 +361,8 @@ export class Node extends GraphPrimitive {
         allowNegativeValues: this.allowNegativeValues,
         valueDefinedSemiQuantitatively: this.valueDefinedSemiQuantitatively,
         frames: _.clone(this.frames),
-        combineMethod: this.combineMethod
+        combineMethod: this.combineMethod,
+        image: this.image
```
knowuh, line 365: I am wondering why we didn't an `image` key in the export format before this.

I guess I also thought that the node was supposed to get its image from the `palettItem`.  But that might be ancient history now.

knowuh (approved):
Looks good to me.

I had a question about how the nodes `image` and `palletItem` relate, if at all.  But its just because I have been away from this project for so long.

The code looks good, thanks!

knowuh (commented):
Looks good to me

PR #680, opened 2021-06-04
Updated countGraphsWithMultiPaths algorithm with tests using Dan's test cases [#177893493]

dougmartin:


knowuh (approved):
Looks good.  

Do we need a test case for > 1 multi-path graphs in the system?  I saw tests for 0 and 1, but didn't see any for multiple multi-paths.

I am not sure we need it, just pointing it out.

dougmartin:
Good point.  I think I can add a test pretty quickly that merges the four new test data files into one file.

dougmartin:
@knowuh I've added a test that combines all four test cases - https://github.com/concord-consortium/building-models/pull/680/commits/902c27129bcb7908481682f0b24ef81cc1460b30

PR #681, opened 2021-06-08
Phase 1: New visualization and relationship setting for collectors and flows

dougmartin:
- Add v-shaped cut out to end of pipe to have a "tail" to the pipe arrow.
- Change add-to relationship to pipe pointing from variable to collector.
- Change subtract-from relationship to pipe pointing from collector to variable


knowuh (approved):
Looks good.

Seems like `_createTransferArrowNotch` is not used at the moment? 

The SVG that function generates  looks more like a star to me when drawn. See: https://codepen.io/knowuh/pen/abJQMzo

dougmartin:
This branch was orphaned when the phase 4 branch was rebased to master.

PR #682, opened 2021-06-08
Phase 2: New visualization and relationship setting for collectors and flows

dougmartin:
If the user selects an "add to" or "subtract from" relationship, then the icon in the variable is changed to the same one we use for the transfer valve.

### DIFF
--- a/src/code/views/node-view.tsx
+++ b/src/code/views/node-view.tsx
@@ -411,8 +411,8 @@ getDefaultProps() {
         );
       } else {
         const outLinks = node.outLinks();
-        const hasAddedToLink = _.filter(outLinks, link => link.relation.formula === "+in").length > 0;
-        const hasSubtractedFromLink = _.filter(outLinks, link => link.relation.formula === "-in").length > 0;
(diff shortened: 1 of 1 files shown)


src/code/views/node-view.tsx
```diff
@@ -410,10 +410,12 @@ getDefaultProps() {
           />
         );
       } else {
+        const outLinks = node.outLinks();
+        const hasAddedToLink = _.filter(outLinks, link => link.relation.formula === "+in").length > 0;
+        const hasSubtractedFromLink = _.filter(outLinks, link => link.relation.formula === "-in").length > 0;
```
kswenson, line 415: Could also use `_.some()` which short-circuits the iteration and eliminates the need for the length check.

dougmartin: @kswenson I switch the code out to use `_.some(...)` after rebasing to the final phase-1 version.

knowuh (approved):
looks good.

dougmartin:
This branch was orphaned when the phase 4 branch was rebased to master.

PR #683, opened 2021-06-08
Phase 3: New visualization and relationship setting for collectors and flows

dougmartin:
- If the user chooses "add to" for a relationship, then use a faucet icon.
- If the user chooses a "subtract from" relationship, then use a drain icon.


knowuh (approved):
Looks good.

PR #684, opened 2021-06-09
WIP: Switch notches to arrows in transfer pipes

dougmartin:


dougmartin:
Deleting - this ended up being the method we adopted in phase1 and it is in that PR now (https://github.com/concord-consortium/building-models/pull/681)

PR #697, opened 2022-01-04
Extracted topology tagger

dougmartin:
This is now stored in another repo and published on npm so that it can be shared with the report-item interactive.


knowuh (approved):
Looks good, thanks

PR #702, opened 2022-01-21
Fix broken mini-graphs/broken initialValue setting [#180624189]

dougmartin:
This fix ensures that the initial value of the node lies within the min/max values of the node.

This bug manifested itself with a saved model that had initial values of two accumulators set at 100000000.  This caused the graphs to not draw correctly as the value was outside of the allowed range.


knowuh (approved):
looks good

PR #704, opened 2022-01-24
Change delete/backspace keyboard event to capture phase [#180950072]

dougmartin:
This ensures that the delete/backspace keyboard event is checked first before other event handlers can block the event from propogating back up.


knowuh (approved):
Looks good to me!

PR #707, opened 2022-01-25
The relationship panel should guide the user to set unselected relationships [#180450641]

dougmartin:
Change double click on a node to select the first unset relationship, or if none, then select the first relationship.


knowuh (approved):
Looks good.

PR #712, opened 2022-01-27
Add focus on startup and tabIndex to root app to ensure top level delete handler is always available [#180950072]

dougmartin:


src/code/views/app-view.tsx
```diff
@@ -121,6 +121,11 @@ export class AppView extends Mixer<AppViewProps, AppViewState> {
     }
 
     window.addEventListener("click", this.handleWindowClick, true);
+
+    // Ensure the app has focus so that the top level window delete key handler is invoked
+    // if an element is selected right after and file load and the delete key is pressed.
```
kswenson, line 126:
```suggestion
    // if an element is selected right after a file load and the delete key is pressed.
```

dougmartin: Fixed in https://github.com/concord-consortium/building-models/pull/712/commits/c024f06fcd5faf2e22d805ee1739cf942de70f34

knowuh (approved):
Looks good, thanks!

PR #713, opened 2022-01-28
Added new flow icon view [#181050970]

dougmartin:
image

TODO:

- figure out better way than a string compare to determine if the icon is a default blank or flow icon


src/code/models/node.ts
```diff
@@ -144,8 +147,8 @@ export class Node extends GraphPrimitive {
...
+          || val.substr(0, BlankImagePrefix.length) === BlankImagePrefix
+          || val.substr(0, FlowImagePrefix.length) === FlowImagePrefix
```
kswenson: Could also use [startsWith](https://caniuse.com/?search=startsWith).

dougmartin: This code was removed entirely and replaced with a migration.

src/code/models/node.ts
```diff
@@ -23,6 +23,9 @@ const SEMIQUANT_MIN = 0;
 const SEMIQUANT_MAX = 100;
 const SEMIQUANT_ACCUMULATOR_MAX = 1000;
 
+const BlankImagePrefix = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACoAAAAqCAYAAADFw8lbAAA";
+const FlowImagePrefix  = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACgAAAAoCAYAAACM/rhtAAA";
+
```
dougmartin, line 28: I need to delete this.  It is not needed anymore.

src/code/models/node.ts
```diff
@@ -77,7 +82,7 @@ export class Node extends GraphPrimitive {
   constructor(nodeSpec, key?, isTransfer?) {
     super(isTransfer ? "Transfer" : "Node");
 
-    let val, val1, val10, val11, val2, val3, val4, val5, val6, val7, val8, val9;
+    let val, val1, val10, val11, val12, val2, val3, val4, val5, val6, val7, val8, val9;
```
knowuh, line 85: Well, 😕 I don't love this. 

I realize this must be the result of some automatic conversion, but ... Seems like we at least need to put a PT story in to fix it.

knowuh (approved):
I made a comment about some legacy conversion code that made my eyes bleed. I think if you have extra time, as a community service you should clean that up, even though it was there before you got to it.

PR #718, opened 2022-02-02
Fix data export upon record [#180562893]

dougmartin:
At some point CODAP started deleting the samples subcollection when the data point column was deleted and the collection had no more attributes.

This fix moves the deleting of the extra column to after the simulation has run so that the collection remains when the column is removed.

This also adds code to re-create the missing subcollection on saved files.


scytacki (approved):
I don't understand all of it, but looks good.

PR #727, opened 2022-03-14
Removed default "New Model" filename [#181175814]

dougmartin:
We used to use "New Model" as the default but this caused
hash(dataOnLoad) !== hash(dataOnSave) which then caused
immediate saves when a new model was loaded.  These
immediate saves caused the linked interactive UI to show
as it compares save times to figure out it content changed.
We opted to remove the default here instead of setting the
default in graphStore.init() so that old saved models, with
null as their filename, would hash the same.


scytacki (approved):
seems fine to me

PR #728, opened 2022-03-14
Fix crash if transfer flow is removed and other variables point to that transfer variable [#181131221]

dougmartin:


knowuh (approved):
Looks good, thanks.

(sorry I didn't see this one earlier!)

PR #730, opened 2022-03-14
Added CloudFront invalidation of all html templates [#181125604]

dougmartin:
Before only index.html was invalidated on S3 deploys.  Now index.html , lara.html, report.html and sagemodeler.html are invalidated.


knowuh (approved):
LGMT Thanks!

PR #731, opened 2022-03-14
Add model type help modal [#159109098]

dougmartin:
Adds a ? icon next to the model which toggles a top level help modal.

There are two issues to resolve:

1. The final text for the modal which then needs to be added to the translation files.

2. The url used in the "Click here" link in the help modal.


knowuh (approved):
Thanks for adding my strings to all the strings files.

I thought that would happen somehow automagically the next time POE edits were merged.

Looks good (second time reviewing)

Thanks

PR #733, opened 2022-03-16
Added collector to node palette [#180450580]

dougmartin:
This also updates the recent changes to the palette and the new node action menu to use the "view" naming convention and the app settings mixin to get the simulation type as it not updating when the simulation type was changed.


knowuh (approved):
Looks good, thanks!

PR #734, opened 2022-03-17
Changed undefined links to be dashed and disabled expanded set of relationships option for diagram only [#177724448]

dougmartin:


dougmartin:
Closed and opened again (https://github.com/concord-consortium/building-models/pull/736) due to GitHub going down.

PR #735, opened 2022-03-17
Fixed second transfer flow variable plotted right on top of the other [#181061656]

dougmartin:
This adds a simple y offset when two nodes have crossing transfer links so that the transfer nodes don't overlay each other.


knowuh (approved):
I already approved once today, but second time is a charm!

thanks

PR #736, opened 2022-03-17
Changed undefined links to be dashed and disabled expanded set of relationships option for diagram only [#177724448]

dougmartin:


knowuh (approved):
Looks good, thanks

PR #737, opened 2022-03-17
181336977 - Quick Action Menu Fix

knowuh:
Dan said that:
> we should not show the flow and collector images

in the quick-actions menu

[#181336977]
https://www.pivotaltracker.com/story/show/181336977


dougmartin (approved):
Looks good :+1:

PR #740, opened 2022-03-21
Fixed basic/expanded relationship radio button [#177724448]

dougmartin:
This fixes a couple of bugs added in a previous commit:

1. Removes direct use of props when setting the initial state to remove React warning
2. Adds a check when changing the simulation type to diagram only to auto change to basic complexity if there are no complex existing relationships

This also removes an unused import found when linting.


knowuh (approved):
LGTM

PR #742, opened 2022-03-22
Update relationship panel to show all relationships pointing into and out of a node. [#164294971]

dougmartin:


knowuh (approved):
Reviewed synchronously

+ LGTM!

PR #743, opened 2022-03-23
Fixed relationship symbol bug introduced after adding them to transfer links [#181337002]

dougmartin:
The commit that added relationship symbols to transfer links broke the placement of the relationship symbols on variable with magnitude links.


knowuh (approved):
LGTM

PR #745, opened 2022-03-23
Switched position of transfer link symbol positions [#181337002]

dougmartin:
The transfer links now have their relationship symbols next to the transfer variables instead of at the ends of the link.

Old:

image

New:

image


dougmartin:
Closing without merging.  This was a misunderstanding on my part.

PR #746, opened 2022-03-23
Fixed z-index of modal background [#159109098]

dougmartin:
This adds the node-z-index constants to the stylus build before the modal styles so that the z-index variable is defined for the background.  With the invalid z-index the background click handler was not being called.


knowuh (approved):
LGTM, thanks!

PR #747, opened 2022-03-23
Add relationship symbols to transfer modifier links [#181337002]

dougmartin:
Before this change the relationship symbols were not shown for transfer modifier links as their magnitude was 0.


knowuh (approved):
Thanks, looks good.

PR #748, opened 2022-03-23
Remove recently added code to remove flow variable links when the node changes [#181336977]

dougmartin:
This reverts a recent change in the user menu.  This also fixes an issue where unconnected nodes converted to flow variables didn't update their image.


knowuh (approved):
Thanks, looks good.

PR #749, opened 2022-03-23
Fix setting of relationships when simulation type changes [#177724448]

dougmartin:
This moves the to automatically update the relationship type based on the model type from the view to the store.


knowuh (approved):
Looks good, thanks

PR #750, opened 2022-03-24
Quick action menu tweaks [#181336977]

dougmartin:
- Only show create graph option for transfer nodes
- Update node image correctly when switching between node types
- Removes flow links when a node is converted to a normal node


knowuh (approved):
Looks good thanks

PR #751, opened 2022-03-27
Fix logic bug when checking if tabbed panel should be hidden.

dougmartin:


dougmartin:
Merging without review - this was found while I was testing another issue and if a simple bug fix.

PR #752, opened 2022-03-29
Removed surrounding border from quick action menu

dougmartin:


dougmartin:
Merging with no review as it is a small css change scoped to one component.

PR #757, opened 2022-03-31
Omnibug bug fix after release [#181740905]

dougmartin:
- Fixed accidental removal of flow links on node change that isn't about node type
- Fixed image update to not update image info if it hasn't changed in the node
- Fixed node setting palette info for node when collector is dragged in.  This was just updating the image but leaving the uuid and useDefaultImage in the incorrect state.
- Updated node view not to "stack" node images if they are already the collector image


lublagg (approved):
Looks good to me!

PR #760, opened 2022-05-18
Reduce size of serialized data [#182142004]

dougmartin:
- Changed resized images from 100x100 to 50x50 as 50x50 is the maximum we ever use
- Changed graph serialization to remove duplicate image in node if it already exists in the palette


dougmartin:
> Code looks clean.
> 
> I have one comment worth looking into. I am not seeing the delegation to `palletteItem.image` happening in the node-view rendering path. I am probably just missing it.

The importer code called on load adds the image to the node if there is a `paletteItem` defined for the node:

https://github.com/concord-consortium/building-models/blob/master/src/code/utils/importer.ts#L46-L48

This code was already is the importer.

PR #784, opened 2023-07-18
fix: SageModeler plugin coming "unattached" from the CODAP canvas [PT-185485076]

dougmartin:
This was caused by the Korean translation of the iframe title parameter which is not translated in any of the other languages.  The title is used when trying to match the standalone parameter and thus can't be translated.

The code has been updated to not use the translation.


pjanik (approved):
Looks good 👍

PR #786, opened 2023-07-19
fix: Prevent transfer node from connecting to additional collectors [PT-184163904]

dougmartin:
This both cleans up the add link logic and adds a new condition to the logic to prevent the user from connecting to/from a transfer node to an "outside" collector, that is, a collector that isn't the source or target of the transfer link.


pjanik (approved):
Looks good 👍

PR #792, opened 2024-01-17
fix: Switch to SageModeler Google Drive project [PT-186851280]

dougmartin:
This was using a Google project we no longer has access to and thus can't enable the Google picker on.

### DIFF
--- a/src/assets/sage.html
+++ b/src/assets/sage.html
@@ -79,8 +79,9 @@
         },
         {
           "name": "googleDrive",
-          "apiKey": "AIzaSyAUobrEXqtbZHBvr24tamdE6JxmPYTRPEA",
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 LGTM -- I didn't verify the ids, but I trust you on those. 😉

dougmartin:
@kswenson btw, this is for an unrelated issue to the current Google drive hotfix.  We like to keep building-models on the same package number as  sage-modeler-site but the old settings were causing a build break on building-models GitHub actions because it didn't have the new appId parameter.

dougmartin:
I've just updated the SageModeler client authorized domain and tested this branch and the Google provider works.

